PR #1, opened 2020-06-25
Implement materialize.css and add navbar

emilyzhaoo:


emilyzhaoo:
@googlebot I signed it!

PR #5, opened 2020-06-30
Display search results in cards

emilyzhaoo:


planet/src/main/webapp/places.js
```diff
@@ -101,27 +100,37 @@ function initMap() {
...
-/** List results of nearby search */
+/** List results of nearby search on page*/
 function listResults(results) {
     const places = document.getElementById('places');
 
+    // Displays search results on page
     for (var i = 0; i < results.length; i++) {
-        places.innerText = places.innerText + <br /> + results[i]; 
+        // Create materialize card for each result by creating dom element in HTML
+        var div1 = document.createElement("div");
```
azhang72: nit: s/var/let.

azhang72 (approved):
lgtm with a nit

planet/src/main/webapp/places.js
```diff
@@ -101,27 +100,37 @@ function initMap() {
...
+/** List results of nearby search on page*/
 function listResults(results) {
     const places = document.getElementById('places');
 
-    for (var i = 0; i < results.length; i++) {
-        places.innerText = places.innerText + <br /> + results[i]; 
+    // Displays search results on page
+    for (let i = 0; i < results.length; i++) {
+        // Create materialize card for each result by creating dom element in HTML
+        let div1 = document.createElement("div");
```
EkkiD, line 112: Why do we need 3 divs? Is there materialize documentation we can link to here?

EkkiD (approved):
LGTM with nit

planet/src/main/webapp/places.js
emilyzhaoo: Is there a way to do that? I manually created the materialize card (which required 3 divs)

EkkiD: I meant a link in the comment to the documentation that explains what the 3 divs are. Maybe this one? https://materializecss.com/cards.html#horizontal

PR #6, opened 2020-07-02
Sign in form

nabil-dbz:
This pull request adds the sign in form to the web page. The user can get logged in by entering their email and password!

### DIFF
--- a/planet/src/main/webapp/js/authentication.js
+++ b/planet/src/main/webapp/js/authentication.js
@@ -12,8 +12,8 @@ const firebaseConfig = {
 
 firebase.initializeApp(firebaseConfig);
 
-function resetForm() {
-    const myFormInputs = document.getElementsByClassName('input-sign-up');
+function resetForm(elementsClass) {
+    const myFormInputs = document.getElementsByClassName(elementsClass);
     if (myFormInputs === null) {
         return;
     }
@@ -22,7 +22,7 @@ function resetForm() {
     }
 }
 
-function submitForm() {
+function signUp() {
     const password = document.getElementById('password').value;
     const passwordConfirmation = document.getElementById('repeat-password').value;
     const email = document.getElementById('email').value;
@@ -31,6 +31,17 @@ function submitForm() {
         return;
     }
     firebase.auth().createUserWithEmailAndPassword(email, password).catch(function(error) {
+        // Handle Errors here.
+        var errorCode = error.code;
+        var errorMessage = error.message;
+        // ...
+    });
+}
+
+function signIn() {
+    const password = document.getElementById('password-in').value;
+    const email = document.getElementById('email-in').value;
+    firebase.auth().signInWithEmailAndPassword(email, password).catch(function(error) {
         // Handle Errors here.
         var errorCode = error.code;
         var errorMessage = error.message;
--- a/planet/src/main/webapp/js/authentication.js
+++ b/planet/src/main/webapp/js/authentication.js
@@ -60,7 +60,7 @@ function signUp() {
     }).catch(function(error) {
         // Handle Errors here.
         console.log(error);
-        // ...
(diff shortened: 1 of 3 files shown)


zhoua0805 (approved):

planet/src/main/webapp/js/authentication.js
```diff
@@ -22,18 +42,35 @@ function resetForm() {
...
+function signIn() {
+    const password = document.getElementById('password-in').value;
+    const email = document.getElementById('email-in').value;
+    resetForm('input-sign-in');
+    firebase.auth().signInWithEmailAndPassword(email, password).catch(function(error) {
         // Handle Errors here.
-        var errorCode = error.code;
-        var errorMessage = error.message;
+        console.log(error);
         // ...
```
emilyzhaoo: nit: Remove unused comment

emilyzhaoo (approved):

planet/src/main/webapp/js/authentication.js
```diff
@@ -22,18 +42,35 @@ function resetForm() {
...
-    firebase.auth().createUserWithEmailAndPassword(email, password).catch(function(error) {
+    firebase.auth().createUserWithEmailAndPassword(email, password)
+    .then(function() {
+        return firebase.auth().currentUser.updateProfile({
+            displayName: displayName,
+        })
+    }).catch(function(error) {
+        // Handle Errors here.
+        console.log(error);
+        // ...
```
EkkiD: For errors that require the user to take action, we should tell the user about them. An alert is probably good enough for the prototype.

https://firebase.google.com/docs/reference/js/firebase.auth.Auth#createuserwithemailandpassword
https://firebase.google.com/docs/reference/js/firebase.auth.Auth#signinwithemailandpassword

EkkiD (changes requested):

planet/src/main/webapp/js/authentication.js
nabil-dbz: Okay! I'll add that right away

planet/src/main/webapp/js/authentication.js
```diff
@@ -22,18 +42,35 @@ function resetForm() {
...
-    firebase.auth().createUserWithEmailAndPassword(email, password).catch(function(error) {
+    firebase.auth().createUserWithEmailAndPassword(email, password)
+    .then(function() {
+        return firebase.auth().currentUser.updateProfile({
+            displayName: displayName,
+        })
+    }).catch(function(error) {
+        // Handle Errors here.
+        console.log(error);
+        alert('Create account with Email and Password failed!');
```
EkkiD: different errors should be handled differently. In particular, if there is an action the user can take to fix the error, we should display a message that tells them what to do.

```suggestion
    }).catch(function(error) {
        console.log(error);
       if (error.code === 'auth/email-already-in-use') {
          alert('There is already an account with this email address. Try signing in instead.');
       } else if (error.code === 'auth/invalid-email') {
         alert('Invalid email address. Double check the spelling.');
       } else if (....) {
        ....
       } else {
            // No specific action the user can take. Show a generic message.
           alert('Create account with Email and Password failed!');
       }
```

EkkiD (changes requested):

planet/src/main/webapp/js/authentication.js
EkkiD: See https://firebase.google.com/docs/reference/js/firebase.auth.Auth#error-codes_3 and  https://firebase.google.com/docs/reference/js/firebase.auth.Auth#error-codes_12  for the possible error codes

nabil-dbz: Got it! Fixed

EkkiD (commented):
LGTM

EkkiD (approved):
LGTM

PR #7, opened 2020-07-02
Create an itinerary in the order of the events

zhoua0805:
- Create an itinerary that schedules events in the order that the users add them in 
- For the prototype, set all travelling times to be 15 minutes
- Next step: allow users to input hotel location (starting point)

### DIFF
--- a/planet/src/main/java/com/google/planet/data/Event.java
+++ b/planet/src/main/java/com/google/planet/data/Event.java
@@ -51,7 +51,7 @@ public final class Event {
     /**
     * A comparator for sorting events by the duration of their opening hours, in ascending order.
     */
-    public static final Comparator<Event> orderByOpeningHours = new Comparator<Event>() {
+    public static final Comparator<Event> OrderByOpeningHours = new Comparator<Event>() {
         @Override
         public int compare(Event a, Event b) {
         return Long.compare(a.openingHours.duration(), b.openingHours.duration());
--- a/planet/src/main/java/com/google/planet/data/Event.java
+++ b/planet/src/main/java/com/google/planet/data/Event.java
@@ -14,6 +14,8 @@
 
 package com.google.planet.data;
 
+import java.util.Comparator;
+
 public final class Event {
     private final long id;
     private final String name; 
@@ -45,5 +47,15 @@ public final class Event {
     public TimeRange getOpeningHours(){
         return this.openingHours;
     }
+
+    /**
+    * A comparator for sorting events by the duration of their opening hours, in ascending order.
+    */
+    public static final Comparator<Event> ORDER_BY_OPENING_HOURS = new Comparator<Event>() {
+        @Override
+        public int compare(Event a, Event b) {
+        return Long.compare(a.openingHours.duration(), b.openingHours.duration());
+        }
+    };
 }
 
--- a/planet/src/main/java/com/google/planet/data/Event.java
+++ b/planet/src/main/java/com/google/planet/data/Event.java
@@ -51,7 +51,7 @@ public final class Event {
     /**
     * A comparator for sorting events by the duration of their opening hours, in ascending order.
     */
-    public static final Comparator<Event> ORDER_BY_OPENING_HOURS = new Comparator<Event>() {
+    public static final Comparator<Event> orderByOpeningHours = new Comparator<Event>() {
         @Override
         public int compare(Event a, Event b) {
         return Long.compare(a.openingHours.duration(), b.openingHours.duration());
--- a/planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
+++ b/planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
@@ -24,7 +24,7 @@ public final class ItineraryGenerator {
 
         // For the prototype, just find the shortest opening hours and 
         // schedule all events within that time range
-        Collections.sort(events, Event.orderByOpeningHours);
+        Collections.sort(events, Event.OrderByOpeningHours);
         int START = events.get(0).getOpeningHours().start();
         int END = events.get(0).getOpeningHours().end();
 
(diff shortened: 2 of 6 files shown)


planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
```diff
@@ -0,0 +1,82 @@
...
+                }else{
+                    travelTimeGraph[i][j] = 15;
+                }
+            }
+        }
+        return travelTimeGraph;
+    }
+
+    // Function that creates an itinerary by scheduling each event in order.
+    public List<ItineraryItem> scheduleItineraryInOrder(List<Event> events, int START, int END,
```
nabil-dbz, line 55: The parameters names should be written in lowerCamelCase (start, end) according to the Google Java Style Guide

planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
```diff
@@ -0,0 +1,82 @@
...
+            }
+        }
+        return travelTimeGraph;
+    }
+
+    // Function that creates an itinerary by scheduling each event in order.
+    public List<ItineraryItem> scheduleItineraryInOrder(List<Event> events, int START, int END, 
+                                                        int[][] travelTimeGraph) {
+        List<ItineraryItem> items = new ArrayList<>();
+        int index = START;
```
nabil-dbz: I think that the variable name index is too general. In this case, it keeps track of the timing, right?

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
```diff
@@ -0,0 +1,72 @@
...
+            double duration = (double) entity.getProperty("duration");
+            int openingTime = Math.toIntExact((long)entity.getProperty("openingTime"));
+            int closingTime = Math.toIntExact((long)entity.getProperty("closingTime"));
+            String listName = (String) entity.getProperty("listName");
+            String userId = (String) entity.getProperty("userId");
+            Event event = new Event(id, name, address, duration, 
+                TimeRange.fromStartEnd(openingTime, closingTime), listName, userId);
+            events.add(event);
+        }
+
```
nabil-dbz: Big part of this doPost is exactly the same as the doGet function of the EventServlet, Is it possible to avoid code duplication in this case?

planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
nabil-dbz: The parameters names should be written in lowerCamelCase (start, end) according to the Google Java Style Guide

zhoua0805: Okay! fixed

planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
zhoua0805: Yep! I changed the naming.

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
zhoua0805: Hmmm I searched up sending responses from one servlet to another. Although it's possible, since eventservlet is run everytime the page loads but itinerary servlet is only run when the user clicks generate itinerary button, I'm not sure if sending requests between servlets are the best option

nabil-dbz: Is it possible to get the current events' data from the client side since they already on the DOM?

nabil-dbz (approved):
LGTM!

planet/src/main/java/com/google/planet/data/Event.java
```diff
@@ -45,5 +47,15 @@ public String getAddress(){
     public TimeRange getOpeningHours(){
         return this.openingHours;
     }
+
+    /**
+    * A comparator for sorting events by the duration of their opening hours, in ascending order.
+    */
+    public static final Comparator<Event> ORDER_BY_OPENING_HOURS = new Comparator<Event>() {
```
azhang72: The comparator only compares. It doesn't order the Events or sort them. Consider rename it to something like EventComparator.
I'm not familiar with Java Comparators, but I assume ORDER_BY_OPENING_HOURS is a class. If so, use PascalCase.

planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
```diff
@@ -17,10 +17,64 @@
 import java.util.List;
 import java.util.Arrays;
 import java.util.ArrayList;
+import java.util.Collections;
 
 public final class ItineraryGenerator {
     public List<ItineraryItem> generateItinerary(List<Event> events) {
```
azhang72, line 23: Is it necessary for all these methods to be public? It seems that you are only testing generateItinerary(), it'd make sense to mark everything else private.

azhang72 (changes requested):

planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
zhoua0805: Okay!

planet/src/main/java/com/google/planet/data/Event.java
zhoua0805: In the example from week 5, the same convention was used: 
image
My understanding is that since comparators are only used for sorting, it makes sense to name them as SortBySomething? I also did a quick search on some examples online and changed the format to PascalCase.:D

azhang72: Ah okay. It's hard to argue when they set up the code like that.

azhang72 (approved):

PR #8, opened 2020-07-02
Add preset search options as buttons

emilyzhaoo:
This PR adds preset search options as buttons for the user. When clicked, the associated keyword will appear in the search box.

### DIFF
--- a/planet/src/main/webapp/places.css
+++ b/planet/src/main/webapp/places.css
@@ -12,10 +12,10 @@ header {
     padding:0; 
 }
 
-h3 {
-    margin: 10px !important;
-    font-size: 30px !important;
-    padding:10px !important;
+#greeting {
+    margin: 0.86vh;
+    font-size: 2.6vh;
+    padding:0.86vh;
 }
 
 #description {
@@ -46,7 +46,7 @@ h3 {
     margin: 0;
     height: 80%;
     overflow-y: scroll;
-    padding-right: 15px;
+    padding-right: 0.87vw;
 }
 
 .row {
@@ -60,15 +60,15 @@ h3 {
 }
 
 .card.horizontal {
-    margin: 10px;
+    margin: 0.86vh;
 }
 
 #button-bar {
-    padding: 10px;
+    padding: 0.86vh;
     text-align: center;
 }
 
 .btn-floating {
-    margin: 10px; 
-    transition: background-color .3s;
+    margin: 0.86vh; 
+    transition: background-color 0.3s;
 }
--- a/planet/src/main/webapp/places.css
+++ b/planet/src/main/webapp/places.css
@@ -12,6 +12,12 @@ header {
     padding:0; 
 }
 
+h3 {
+    margin: 10px !important;
+    font-size: 30px !important;
+    padding:10px !important;
+}
+
 #description {
     font-family: Roboto;
     font-size: 15px;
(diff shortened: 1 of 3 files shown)


planet/src/main/webapp/places.css
```diff
@@ -12,6 +12,12 @@ header {
     padding:0; 
 }
 
+h3 {
+    margin: 10px !important;
+    font-size: 30px !important;
+    padding:10px !important;
+}
```
nabil-dbz: It's better to avoid using "important". This part of the code would oblige all the "h3" elements including those of your teammates to follow this styling. You may wanna add them to a class!

planet/src/main/webapp/places.html
```diff
@@ -25,12 +25,19 @@
 </header>
 
 <body >
-    <div id="column" style="width:75%;" >
-        <input id="pac-input" class="controls " style="width:50%; background:white;" type="text" placeholder="Search Location">
+    <div id="column" style="width:75%;">
+        <input id="pac-input" class="controls " style="width:50%; background:white; margin:5px; padding-left:10px;" type="text" placeholder="Search Location">
```
nabil-dbz: nit: I think it's better to use "vw" or "vh" instead of "px"

planet/src/main/webapp/places.js
```diff
@@ -102,39 +108,46 @@ function initMap() {
...
-        let div1 = document.createElement("div");
-        div1.classList.add("card"); 
-        div1.classList.add("horizontal"); 
+        let div1 = document.createElement('div');
+        div1.classList.add('card'); 
+        div1.classList.add('horizontal'); 
 
-        let div2 = document.createElement("div");
+        let div2 = document.createElement('div');
         div2.classList.add("card-stacked");
```
nabil-dbz: nit: To keep the harmony of the code, single quotes for "card-stacked"

planet/src/main/webapp/places.js
```diff
@@ -48,13 +41,26 @@ function initMap() {
...
           handleLocationError(false, infoWindow, map.getCenter());
         }
+    updateSearch();
+}
+
+/** Create new markers and display new search */
+function updateSearch() {    
+    let input = document.getElementById('pac-input');
+    let searchBox = new google.maps.places.SearchBox(input);
+    map.controls[google.maps.ControlPosition.TOP_LEFT].push(input);
```
EkkiD, line 54: What does this line do? It might warrant a comment explaining what's happening

planet/src/main/webapp/places.js
```diff
@@ -102,39 +108,46 @@ function initMap() {
     });
 }
 
+function handleLocationError(browserHasGeolocation, infoWindow, pos) {
```
EkkiD, line 114: It looks like we only support the case where the user has location enabled. Is there a way we can gracefully degrade if they have it disabled? Like allowing them to search?

EkkiD (changes requested):

planet/src/main/webapp/places.js
emilyzhaoo: The user is still able to search even if they don't have location allowed

emilyzhaoo: Right now, there is an error message that pops up as an info window marker on the map - when that feature is improved I will make changes accordingly but we still aren't sure what to do with the user location

EkkiD (approved):
LGTM

nabil-dbz (approved):
LGTM!

PR #9, opened 2020-07-03
Add hotel feature to itinerary

zhoua0805:
**Note:** this PR comes after PR #7 

- allow users to input hotel location in the front end
- store hotel location in sessionStorage (only disappear if the user closes the browser)
- add hotel as an event with duration of 0 in the back end
- generate an itinerary with hotel as the starting point

### DIFF
--- a/planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
+++ b/planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
@@ -20,7 +20,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 
 public final class ItineraryGenerator {
-    public List<ItineraryItem> generateItinerary(List<Event> events) { 
+    public List<ItineraryItem> generateItinerary(List<Event> events, Event hotel) { 
 
         // For the prototype, just find the shortest opening hours and 
         // schedule all events within that time range
@@ -28,6 +28,8 @@ public final class ItineraryGenerator {
         int START = events.get(0).getOpeningHours().start();
         int END = events.get(0).getOpeningHours().end();
 
+        // Add hotel as the "first" event
+        events.add(0, hotel);
         int [][] travelTimeGraph = getTravelTimeGraph(events);
         List<ItineraryItem> items = scheduleItineraryInOrder(events, START, END, travelTimeGraph);
         return items;
@@ -56,9 +58,11 @@ public final class ItineraryGenerator {
                                                         int[][] travelTimeGraph) {
         List<ItineraryItem> items = new ArrayList<>();
         int start = openningTime; 
-        for (int i = 0; i < events.size(); i++){
+        for (int i = 1; i < events.size(); i++){  //Skip the first event (hotel)
             Event event = events.get(i);
 
+            // Update start time with travelling time.
+            start += travelTimeGraph[i-1][i];
             // Add the event as an itinerary item if the remaining available time is longer than the
             // event duration.
             if (start + event.getDurationInMinutes() <= endingTime) {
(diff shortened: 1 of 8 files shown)


nabil-dbz (approved):

planet/src/main/webapp/js/itinerary.js
```diff
@@ -60,8 +80,14 @@ function createItinerary(items) {
     itineraryContainer.innerHTML = '';
 
     items.forEach((item) => {
-        itineraryContainer.innerHTML += '<li>' + item.name + ', ' + item.address + ', ' + timeToString(item.when.start) +
-            ' - ' + timeToString(item.when.end) + '</li>';
+        // Only show one time stamp for hotel.
```
EkkiD: this comment doesn't match the if statement. This doesn't need to only apply to hotels, it applies to any zero-length events.

planet/src/main/webapp/js/itinerary.js
```diff
@@ -7,6 +7,20 @@ function closeForm() {
     document.getElementById('add-event').style.display = 'none';
 }
 
+function handleHotelChange() {
+    if (typeof(Storage) !== "undefined") {
+        sessionStorage.hotel = document.getElementById('hotel-address').value;
```
EkkiD: use getItem and setItem to correspond to spec: https://www.w3schools.com/jsref/prop_win_sessionstorage.asp

planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
```diff
@@ -56,7 +58,7 @@
                                                         int[][] travelTimeGraph) {
         List<ItineraryItem> items = new ArrayList<>();
         int start = openningTime; 
-        for (int i = 0; i < events.size(); i++){
+        for (int i = 0; i < events.size(); i++){
```
EkkiD: nit: remove trailing whitespace

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
```diff
@@ -45,6 +45,16 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
 
+        // Create an event called "hotel" with duration of 0
+        String hotelAddress = request.getParameter("hotel-address");
+        Event hotel = new Event(123, 
+                                "Hotel", 
+                                hotelAddress, 
+                                0, 
+                                TimeRange.WHOLE_DAY, 
+                                "listName-none", 
+                                "userId-none");
```
EkkiD: rather than putting garbage data into this API, these parameters should be optional.

Suggest making a separate constructor for Event that omits ID, list name, and user id.

EkkiD (changes requested):

planet/src/main/webapp/js/itinerary.js
zhoua0805: fixed!

planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
zhoua0805: Okay!

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
zhoua0805: Ah Okay!

planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
zhoua0805: Okay!

planet/src/main/webapp/js/itinerary.js
zhoua0805: Done!

EkkiD (commented):
LGTM

EkkiD (approved):

PR #10, opened 2020-07-06
Sign in with accounts

nabil-dbz:
This pull request add to the project the option to sign in with Google, Facebook or Github account.

### DIFF
--- a/planet/src/main/webapp/js/authentication.js
+++ b/planet/src/main/webapp/js/authentication.js
@@ -77,6 +77,20 @@ function signIn() {
 
 function signInWithGoogle() {
     const provider = new firebase.auth.GoogleAuthProvider();
+    signInWithProvider(provider);
+}
+
+function signInWithFacebook() {
+    const provider = new firebase.auth.FacebookAuthProvider();
+    signInWithProvider(provider);
+}
+
+function signInWithGithub() {
+    const provider = new firebase.auth.GithubAuthProvider();
+    signInWithProvider(provider);
+}
+
+function signInWithProvider(provider) {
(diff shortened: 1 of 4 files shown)


planet/src/main/webapp/js/authentication.js
```diff
@@ -74,3 +74,67 @@ function signIn() {
         // ...
     });
 }
+
+function signInWithGoogle() {
+    const provider = new firebase.auth.GoogleAuthProvider();
```
emilyzhaoo: Maybe try renaming these consts to be more specific, const should always be in capital case ex: GOOGLE_PROVIDER

planet/src/main/webapp/js/authentication.js
```diff
@@ -74,3 +74,67 @@ function signIn() {
...
+
+function signInWithGithub() {
+    const provider = new firebase.auth.GithubAuthProvider();
+    signInWithProvider(provider);
+}
+
+function signInWithProvider(provider) {
+    firebase.auth().signInWithPopup(provider).then(function(result) {
+        // The signed-in user info.
+        var user = result.user;
```
emilyzhaoo: nit: Use let instead of var

emilyzhaoo (approved):

planet/src/main/webapp/js/authentication.js
nabil-dbz: Fixed!

planet/src/main/webapp/js/authentication.js
nabil-dbz: Fixed!

azhang72 (approved):

PR #12, opened 2020-07-06
Database set up

nabil-dbz:
This pull request makes use of the Firebase Realtime Database to add the users to the database.

### DIFF
--- a/planet/src/main/webapp/js/authentication.js
+++ b/planet/src/main/webapp/js/authentication.js
@@ -25,6 +25,7 @@ const firebaseConfig = {
 
 firebase.initializeApp(firebaseConfig);
 const database = firebase.database();
+// An example of how to retrieve data from the database (demos puposes)
 database.ref('users').on('value', function(snapshot) {
     console.log(snapshot.val());
     updateListOfUsers(snapshot.val());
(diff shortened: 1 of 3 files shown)


zhoua0805 (approved):
LGTM

planet/src/main/webapp/js/authentication.js
```diff
@@ -24,6 +24,11 @@ const firebaseConfig = {
 };
 
 firebase.initializeApp(firebaseConfig);
+const database = firebase.database();
+database.ref('users').on('value', function(snapshot) {
+    console.log(snapshot.val());
+    updateListOfUsers(snapshot.val());
+});
```
EkkiD, line 32: This looks like it's retrieving a list of all users for display. That won't scale beyond a few dozen users. What is the plan to support a larger number of users?

planet/src/main/webapp/js/authentication.js
```diff
@@ -48,13 +64,20 @@ function signUp() {
...
         return;
     }
     firebase.auth().createUserWithEmailAndPassword(email, password)
     .then(function() {
-        return firebase.auth().currentUser.updateProfile({
+        const user = firebase.auth().currentUser;
+        database.ref('users/' + user.uid).set({
+            name: displayName,
+            email: email,
+            phoneNumber: phoneNumber,
```
EkkiD, line 80: based on the documentation, it looks like this data is shared in realtime with all users, so this effectively shares the email and phone number of each user on the site with each other user.

Am I interpreting that correctly? If so, that won't satisfy our privacy requirements. We'll need to look for an alternate design.

EkkiD (changes requested):

planet/src/main/webapp/js/authentication.js
nabil-dbz: This is just an example of how to write data in the database. It will be removed later!

planet/src/main/webapp/js/authentication.js
nabil-dbz: This is just an example of how to retrieve data in the database. It will be removed later!

EkkiD (approved):
LGTM

PR #13, opened 2020-07-06
Allow users to reorder the events cards

zhoua0805:
**Note**: this PR comes after PR #9 

- Use jQuery sortable plugin to enable draggable (reorder) feature
- Add property order to the event class and store it in datastore
- **TODO**: since it seems like we'll be switching to using Firebase on the client side, I didn't connect the reordering feature between client and server (firebase will do everything on the client side). Data migration from datastore to firebase will happen later today - tomorrow.

### DIFF
--- a/planet/src/main/java/com/google/planet/servlets/EventServlet.java
+++ b/planet/src/main/java/com/google/planet/servlets/EventServlet.java
@@ -78,10 +78,7 @@ public class EventServlet extends HttpServlet {
         // Get the count of the current list (will need to add more filters, such as userid, listName, etc. later)
         DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
         PreparedQuery entityStats = datastore.prepare(new Query("Event"));
-        long numOfEvents = 0;
-        for (Entity entity : entityStats.asIterable()) {
-            numOfEvents += 1;
-        }
+        int numOfEvents = entityStats.countEntities();
         long order = numOfEvents + 1;
 
         String listName = "current";
--- a/planet/src/main/java/com/google/planet/servlets/EventServlet.java
+++ b/planet/src/main/java/com/google/planet/servlets/EventServlet.java
@@ -39,7 +39,7 @@ import javax.servlet.http.HttpServletResponse;
 public class EventServlet extends HttpServlet {
     @Override
     public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-        Query query = new Query("Event");
+        Query query = new Query("Event").addSort("order", SortDirection.ASCENDING);
         DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
         PreparedQuery results = datastore.prepare(query);
 
@@ -51,11 +51,12 @@ public class EventServlet extends HttpServlet {
             double duration = (double) entity.getProperty("duration");
             int openingTime = Math.toIntExact((long)entity.getProperty("openingTime"));
             int closingTime = Math.toIntExact((long)entity.getProperty("closingTime"));
+            long order = (long)entity.getProperty("order");
             String listName = (String) entity.getProperty("listName");
             String userId = (String) entity.getProperty("userId");
 
             Event event = new Event(id, name, address, duration, 
-                TimeRange.fromStartEnd(openingTime, closingTime), listName, userId);
+                TimeRange.fromStartEnd(openingTime, closingTime), order, listName, userId);
             events.add(event);
         }
 
@@ -74,6 +75,15 @@ public class EventServlet extends HttpServlet {
         int openingTime = TimeRange.getTimeInMinutes(8,0);
         int closingTime = TimeRange.getTimeInMinutes(17,0);
 
+        // Get the count of the current list (will need to add more filters, such as userid, listName, etc. later)
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        PreparedQuery entityStats = datastore.prepare(new Query("Event"));
+        long numOfEvents = 0;
+        for (Entity entity : entityStats.asIterable()) {
+            numOfEvents += 1;
+        }
+        long order = numOfEvents + 1;
+
         String listName = "current";
         String userId = "123";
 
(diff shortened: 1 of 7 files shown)


planet/src/main/java/com/google/planet/servlets/EventServlet.java
```diff
@@ -74,6 +75,15 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
         int openingTime = TimeRange.getTimeInMinutes(8,0);
         int closingTime = TimeRange.getTimeInMinutes(17,0);
 
+        // Get the count of the current list (will need to add more filters, such as userid, listName, etc. later)
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        PreparedQuery entityStats = datastore.prepare(new Query("Event"));
+        long numOfEvents = 0;
+        for (Entity entity : entityStats.asIterable()) {
+            numOfEvents += 1;
+        }
```
nabil-dbz: I think there is a way to retrieve the number of entities without having to loop through all the elements. Check this: https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/PreparedQuery.html#countEntities-com.google.appengine.api.datastore.FetchOptions-

planet/src/main/webapp/js/itinerary.js
```diff
@@ -7,6 +7,20 @@ function closeForm() {
     document.getElementById('add-event').style.display = 'none';
 }
 
+function handleHotelChange() {
+    if (typeof(Storage) !== "undefined") {
```
nabil-dbz: nit: single quotes for undefined

planet/src/main/webapp/js/itinerary.js
```diff
@@ -20,7 +34,7 @@ async function renderEvents() {
 
 function createEventElement(id, name, address, duration) {
     const eventElement = document.createElement('div');
-    eventElement.setAttribute("class", "card");
+    eventElement.setAttribute("class", "card event");
```
nabil-dbz: nit: single quotes

planet/src/main/webapp/js/itinerary.js
```diff
@@ -41,13 +55,19 @@ async function deleteEvent(id) {
 }
 
 async function generateItinerary() {
-    const itineraryResponse = await fetch('/generate-itinerary', {method: 'POST'});
+    if (!sessionStorage.hotel) {
+        alert('Please input a valid hotel address');
+        return;
+    }
+    const params = new URLSearchParams();
+    params.append("hotel-address", sessionStorage.hotel);
```
nabil-dbz: nit: single quotes

planet/src/main/java/com/google/planet/servlets/EventServlet.java
zhoua0805: Will try! I saw another method but it can only be used in production mode. Hopefully this one works

planet/src/main/webapp/js/itinerary.js
```diff
@@ -7,6 +7,20 @@ function closeForm() {
     document.getElementById('add-event').style.display = 'none';
 }
 
+function handleHotelChange() {
+    if (typeof(Storage) !== "undefined") {
+        sessionStorage.setItem('hotel', document.getElementById('hotel-address').value);
+    } else {
+        alert ('Please update your browser');
```
emilyzhaoo, line 14: Is there an alternative solution for when one's browser doesn't work so the use can still use the itinerary builder?

planet/src/main/webapp/js/itinerary.js
```diff
@@ -7,6 +7,20 @@ function closeForm() {
...
 
+function handleHotelChange() {
+    if (typeof(Storage) !== "undefined") {
+        sessionStorage.setItem('hotel', document.getElementById('hotel-address').value);
+    } else {
+        alert ('Please update your browser'); 
+    }
+}
+
+function renderHotel() {
```
emilyzhaoo: nit: as we discussed previously, I think we should rename all the "hotel" variables to a starting point (the user might not start at a hotel) ex: startLocation

planet/src/main/webapp/js/itinerary.js
```diff
@@ -71,3 +97,17 @@ document.addEventListener('DOMContentLoaded', function() {
 });
```
emilyzhaoo: nit: remove extra line

planet/src/test/java/com/google/planet/ItineraryGeneratorTest.java
```diff
@@ -44,20 +44,22 @@ public void setUp() {
     public void scheduleItineraryInOrder() {
```
emilyzhaoo, line 44: maybe add comments for each test that describes what you are testing

emilyzhaoo (approved):

planet/src/main/webapp/js/itinerary.js
zhoua0805: It seems like most browsers will support this except for internet explorer 8 or lower 
image
There are also two alternatives I can think of:
since session storage is just for remembering user's input, I can still allow users to see their itinerary while showing the alert box as a warning. Users will lose their input address every time they go on another page and come back, but the itinerary builder will work fine. 
Another solution is to store it in the database and I will have to figure out a way to distinguish it from the normal events both on the client side and server side. 
Could I address this problem later after I move things to firebase?

planet/src/main/java/com/google/planet/servlets/EventServlet.java
zhoua0805: It worked!

planet/src/test/java/com/google/planet/ItineraryGeneratorTest.java
zhoua0805: Okay!

nabil-dbz (approved):
LGTM

planet/src/main/webapp/js/itinerary.js
zhoua0805: Yes! I will fix that.

planet/src/main/webapp/js/itinerary.js
zhoua0805: okay

EkkiD (approved):
LGTM

PR #15, opened 2020-07-08
Add/delete events using firebase

zhoua0805:
This PR mainly changes itinerary.js to enable data storing using firebase realtime database. More specifically: 
- Render a list of events on the front end. This function listens to database changes and will update automatically after an event is added/deleted/edited
- Add an event to the firebase
- Delete an event
- Re-order the events on the front end and update the "order" property in firebase. The purpose of the re-ordering feature is to give users flexibility to plan their itinerary. When the "generate itinerary" button is clicked, users can choose to generate an itinerary in their preferred order or use our optimized algorithm. 

> 
The data structure used in firebase looks like this: 

         users --> user1  --> name
                          --> email
                          --> currentList --> event 1
                                          --> event 2
                          --> savedList1 --> event 1
                                         --> event 2
                          --> savedList2 ...

TODO: 
- input validation when adding an event
- add success/ failure callbacks
- send the list of events to itinerary servlet
- delete event servlet

### DIFF
--- a/planet/src/main/webapp/itinerary.html
+++ b/planet/src/main/webapp/itinerary.html
@@ -1,60 +1,58 @@
 <!DOCTYPE html>
 <html>
-  <head lang="en">
-    <meta charset="UTF-8">
-    <title>Planet</title>
-    <link rel="stylesheet" href="css/itinerary.css">
+    <head lang="en">
+        <meta charset="UTF-8">
+        <title>Planet</title>
+        <link rel="stylesheet" href="css/itinerary.css">
 
-    <!--Import fonts and icons-->
-    <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
-    <link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Allerta+Stencil" />
-    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
+        <!--Import fonts and icons-->
+        <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
+        <link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Allerta+Stencil" />
+        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
 
-    <!-- Compiled and minified CSS from Materialize -->
-    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/materialize/1.0.0/css/materialize.min.css">
+        <!-- Compiled and minified CSS from Materialize -->
+        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/materialize/1.0.0/css/materialize.min.css">
 
-    <!-- Compiled and minified JavaScript -->
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/materialize/1.0.0/js/materialize.min.js"></script>
+        <!-- Compiled and minified JavaScript -->
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/materialize/1.0.0/js/materialize.min.js"></script>
 
-    <!-- Include all the jQuery libraries -->   
-    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
-    <script src ="https://code.jquery.com/jquery-1.10.2.js"></script>
-    <script src="https://code.jquery.com/ui/1.10.4/jquery-ui.js"></script>
+        <!-- Include all the jQuery libraries -->   
+        <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
+        <script src ="https://code.jquery.com/jquery-1.10.2.js"></script>
+        <script src="https://code.jquery.com/ui/1.10.4/jquery-ui.js"></script>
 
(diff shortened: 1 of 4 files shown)


planet/src/main/webapp/itinerary.html
```diff
@@ -21,35 +21,37 @@
     <script src ="https://code.jquery.com/jquery-1.10.2.js"></script>
     <script src="https://code.jquery.com/ui/1.10.4/jquery-ui.js"></script>
 
-    <!-- Include the JS file -->   
+    <!-- Include the firebase and local JS files -->  
+     <script src="https://www.gstatic.com/firebasejs/4.1.3/firebase.js"></script>
```
azhang72: nit: indentation

planet/src/main/webapp/itinerary.html
```diff
@@ -21,35 +21,37 @@
...
+  <body onload ="renderStartingLocation();">   
       <div id ="content">
         <div id="events-container">
             <a class="btn-floating btn-large" onclick="openForm()"> <i class="material-icons">add</i> </a>
             <p>Starting Address (Hotel/Home): 
                     <input type="text" id="starting-address" name="starting-address" onchange="handleStartingLocationChange()"> </input></p>
             <div id="add-event" class="add-event"> 
                 <a id="close-form" onclick="closeForm()"><i class="material-icons">close</i></a></br>
-                <form action="/update-event" method="POST">   
+                <!-- <form>    -->
```
azhang72: nit: remove unused code

planet/src/main/webapp/itinerary.html
```diff
@@ -21,35 +21,37 @@
...
                 <p>Event Address:
-                    <input type="text" name="add-event-address" id="add-event-address" required></input></p>
+                    <input type="text" name="add-event-address" id="add-event-address"></input></p>
                 <p>Duration (in hours): 
-                    <input type="number" id="add-event-duration" name="add-event-duration" max="12" min="0.5" step = "0.5" required> </input></p>
-                <button class="btn waves-effect waves-light" type="submit">Submit</button></br>
-                </form>
+                    <input type="number" id="add-event-duration" name="add-event-duration" max="12" min="0.5" step = "0.5"> </input></p>
+                <button class="btn waves-effect waves-light" onclick="addEvent();">Submit</button></br>
+                <!-- </form> -->
```
azhang72: nit: remove unused code

planet/src/main/webapp/js/itinerary.js
```diff
@@ -21,37 +39,104 @@ function renderStartingLocation() {
...
+}
+
+// Check authentication status and render the list of events if the user has
+// sign in.
+// Todo: discuss with teammates to decide what to show if user is not signed in.
+firebase.auth().onAuthStateChanged(function(user) {
+  if (user) {
+    renderEvents('currentList');
+  } else {
+     console.log('Please sign in');
```
azhang72: Are you using this for debugging? Users won't typically see this, right?
indentation also looks off.

planet/src/main/webapp/js/itinerary.js
```diff
@@ -96,17 +181,48 @@ document.addEventListener('DOMContentLoaded', function() {
...
-        console.log('order is ' + i + ' for ' + this.innerHTML);
+        const userId = firebase.auth().currentUser.uid;
+        const eventRef = database.ref('users/' + userId +'/currentList/' + this.id);
+        eventRef.update({
+            order: i+1
+        });
     });
 } 
+
+// This is a class representing a time range, it mirrors the TimeRange class in the server
```
azhang72, line 204: nit: move this class to its own file.

azhang72 (changes requested):

planet/src/main/webapp/itinerary.html
zhoua0805: Fixed!

planet/src/main/webapp/itinerary.html
zhoua0805: Okay!

planet/src/main/webapp/itinerary.html
zhoua0805: Okay!

planet/src/main/webapp/js/itinerary.js
zhoua0805: I added a ToDo above this because I think we should prompt users to sign in, either by showing an alert box or redirecting them to the sign in page directly. We also need to decide if users can see the itinerary page at all if they are not signed in. Since we haven't really discussed these ideas yet, I was planning to work on the else statement later.
Thanks for reminding me about the indentation! I just fixed it.

emilyzhaoo (approved):
LGTM

planet/src/main/webapp/js/itinerary.js
zhoua0805: In order to import one class to another class, I need to declare my javascript files as modules instead of standard scripts in html. I tried it but my it won't load my script (if I call a function in the html file, it would say the function is not defined). I did some research and found that each module has their own scope, so the html file won't be able to use functions in the javascript module unless I call all the functions in js or I made every function global. I think both options require many changes that may take very long and was wondering if I can keep the class inside in this case.

Here are the resources I found:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules
https://stackoverflow.com/questions/53630310/use-functions-defined-in-es6-module-directly-in-html
image

EkkiD: I'm interested in exploring this further. There must be some way that we can set this up to allow us to organize our code.

However, I don't think we need to block this PR from proceeding.

EkkiD (approved):
LGTM

azhang72 (approved):

PR #16, opened 2020-07-08
Sign out and Open profile

nabil-dbz:
This pull request adds the sign out feature as well as a small introduction to opening the user's profile.


planet/src/main/webapp/js/authentication.js
```diff
@@ -40,9 +43,26 @@ document.addEventListener('DOMContentLoaded', function() {
...
+    currentUser = firebase.auth().currentUser;
+    if (currentUser !== null) {
+        document.getElementById('profile-button').innerText = currentUser.displayName;
+        document.getElementById('profile-button').style.display = 'block';
+        document.getElementById('sign-out-button').style.display = 'block';
+        document.getElementById('sign-in-button').style.display = 'none';
+    } else {
+        document.getElementById('sign-in-button').style.display = 'block';
+        document.getElementById('profile-button').style.display = 'none';
+        document.getElementById('sign-out-button').style.display = 'none';
```
EkkiD: nit: put these in the same order for consistency

planet/src/main/webapp/js/authentication.js
```diff
@@ -40,9 +43,26 @@ document.addEventListener('DOMContentLoaded', function() {
     loadElement('signup.html', 'sign-up-modal');
 });
 
+function checkUserSignIn() {
+    currentUser = firebase.auth().currentUser;
```
EkkiD, line 47: this is implicitly global. Declare it explicitly as a global for clarity

EkkiD (changes requested):

planet/src/main/java/com/google/planet/servlets/UserServlet.java
```diff
@@ -0,0 +1,40 @@
...
+        // This part is for demos purposes. We will be using RequestDispatcher to forward the request to another JSP file
+        response.getWriter().println("<html>");
+        response.getWriter().println("<head>");
+        response.getWriter().println("</head>");
+        response.getWriter().println("<body>");
+        response.getWriter().println("<h2>Welcome to your profile " + this.userName + "</h2>");
+        response.getWriter().println("<h3>Your email adress: " + this.userEmail + "</h3>");
+        response.getWriter().println("<h3>Your phone number: " + this.userPhone + "</h3>");
+        response.getWriter().println("</body>");
+        response.getWriter().println("</html>");
```
emilyzhaoo: I know this is temporary but if you were planning to create an HTML element in the future as well, there is an easier way to do this in JS to create dom elements in HTML as nodes: https://www.w3schools.com/js/js_htmldom_nodes.asp

emilyzhaoo (approved):

planet/src/main/webapp/js/authentication.js
nabil-dbz: Fixed!

planet/src/main/webapp/js/authentication.js
```diff
@@ -194,3 +234,18 @@ function loadElement(href, elementId) {
...
+        return;
+    }
+    database.ref('users/' + currentUser.uid).once('value').then(function(snapshot) {
+        const params = new URLSearchParams();
+        params.append('email', snapshot.val().email);
+        params.append('name', snapshot.val().name);
+        params.append('phone', snapshot.val().phoneNumber);
+        fetch('user', {method: 'POST', body: params}).then(() => {
+            window.location.href = '/user';
+        });
```
zhoua0805: Is there a way to not go through the servlet at all? I was thinking we can just direct users to the /user page and get the firebase data in its own js file

nabil-dbz: We can do that, but the data in the profile page changes with the current user. We'll need a template to be able to use dynamic HTML or use a Java servlet with a JSP file (which is what I'm doing).
Using a java servlet also complicates things because if I try accessing the page "/user" without going through "places.html", the user data in the servlet will just be "null". I'm trying to find a way to solve that in the profile-page branch.

planet/src/main/java/com/google/planet/servlets/UserServlet.java
EkkiD: I see you've resolved this conversation but I don't see how you've addressed this comment. Please either update the code accordingly or respond here with why that's not feasible.

planet/src/main/java/com/google/planet/servlets/UserServlet.java
```diff
@@ -0,0 +1,40 @@
...
+@WebServlet("/user")
+public class UserServlet extends HttpServlet {
+    private String userName;
+    private String userEmail;
+    private String userPhone;
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response)
+      throws IOException {
+        this.userName = request.getParameter("name");
```
EkkiD: What's the scope of this memory? Does each user get their own instance of this class, or is it shared among users?

EkkiD (changes requested):

planet/src/main/java/com/google/planet/servlets/UserServlet.java
nabil-dbz: Indeed, we can use this approach along with the JSP file idea, but since this PR is more about signing the user out and just opening the profile page, I thought I would address that in another PR in a more structured way.to

planet/src/main/java/com/google/planet/servlets/UserServlet.java
nabil-dbz: This is also something I am working on in another PR. For now, the properties of this class are all static but I still have to further explore my options regarding the issue.

EkkiD: so if I'm understanding correclty, this means that if two users made this request around the same time, User1 might get the profile page for User2, or vice versa?

We'll need to fix that before submitting. The best fix is probably to put the relevant information into the doGet request parameters and to remove the doPost functionality completely. Although an alternative would be to use a map indexed by userId to cache the info

EkkiD (changes requested):

planet/src/main/java/com/google/planet/servlets/UserServlet.java
nabil-dbz: True! I am now working on that on another branch. Should I just remove the openProfile function from this PR and implement it in a more detailed way in my current one?

EkkiD: That's up to you. Either you can edit this PR to make it work for multiple users, or you can remove this feature entirely from this PR and add it in a future PR.

EkkiD (changes requested):

planet/src/main/java/com/google/planet/servlets/UserServlet.java
nabil-dbz: Okay! I just deleted it and I will add it to another PR

EkkiD (approved):
LGTM

PR #17, opened 2020-07-08
Send events to the itinerary servlet

zhoua0805:
- Send the list of events to the itinerary servlet in json format
- Use Gson to convert the json string to a list of event objects in java
- Modify the Event class so that Gson can correctly convert json string to event objects.


planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
```diff
@@ -35,52 +37,33 @@
...
 
-        ItineraryGenerator itineraryGenerator = new ItineraryGenerator();
-        List<ItineraryItem> itinerary = itineraryGenerator.generateItinerary(events, start);
+            ItineraryGenerator itineraryGenerator = new ItineraryGenerator();
+            List<ItineraryItem> itinerary = itineraryGenerator.generateItinerary(events);
 
-        response.setContentType("application/json");
-        String json = new Gson().toJson(itinerary);
-        response.getWriter().println(json);
+            response.setContentType("application/json");
```
emilyzhaoo, line 60: nit: single quotations

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
```diff
@@ -35,52 +37,33 @@
...
+            String json = new Gson().toJson(itinerary);
+            response.getWriter().println(json);
+		}else{
+            response.setContentType("application/json");
+            response.getWriter().println("");
```
emilyzhaoo, line 65: nit: single quotations

planet/src/main/webapp/js/itinerary.js
```diff
@@ -21,47 +49,133 @@ function renderStartingLocation() {
...
+        console.log('Please sign in');
+    }
+});
 
-    events.forEach((event) => {
-        let eventElement = createEventElement (event.id, event.name, 
-            event.address, event.duration);
-        document.getElementById('events').appendChild(eventElement);
+function renderEvents(listName) {
+    const userId = firebase.auth().currentUser.uid;
```
emilyzhaoo, line 94: consider making this a global variable (you instantiate it and declare it the same thing in multiple areas)

emilyzhaoo (approved):

planet/src/main/webapp/js/itinerary.js
azhang72: The functions can be called from different user sessions, right? I think it makes sense to query for uid every time we need it.

zhoua0805: True, I'll leave it as is and see if there's a need to simplify the code in the future

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
zhoua0805: I think in java single quotes and double quotes are different and we should use double quotations for strings

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
```diff
@@ -35,52 +37,33 @@
...
+        BufferedReader br = request.getReader();
+		String eventsJson = "";
 
-        // Create an event called "hotel" with duration of 0
-        String startingAddress = request.getParameter("starting-address");
-        Event start = new Event( "Start", 
-                                startingAddress, 
-                                0);
+		if (br != null){
+			eventsJson = br.readLine();
```
EkkiD: nit: looks like the indentation here is off.

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
```diff
@@ -35,52 +37,33 @@
...
+            ItineraryGenerator itineraryGenerator = new ItineraryGenerator();
+            List<ItineraryItem> itinerary = itineraryGenerator.generateItinerary(events);
 
-        response.setContentType("application/json");
-        String json = new Gson().toJson(itinerary);
-        response.getWriter().println(json);
+            response.setContentType("application/json");
+            String json = new Gson().toJson(itinerary);
+            response.getWriter().println(json);
+		}else{
```
EkkiD: nit: add spaces between the braces and the else: https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace

EkkiD: in what circumstances would the buffered reader be null?

Is that a user input error? Or is it an internal server error? Either way, we should return the corresponding HTTP code: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

planet/src/main/webapp/js/itinerary.js
```diff
@@ -21,47 +49,133 @@ function renderStartingLocation() {
...
+    const eventDuration = document.getElementById('add-event-duration').value;
+    const userId = firebase.auth().currentUser.uid; 
+    
+    // Set every event's opening hours to 8am - 5pm for now
+    const openingTime = TimeRange.getTimeInMinutes(8,0);
+    const closingTime = TimeRange.getTimeInMinutes(17,0);
+
+    const eventListRef = database.ref('users/' + userId + '/currentList');
+    // Get the order number by counting existing events
+    const snap = await eventListRef.once('value');
```
EkkiD: Is 'snap' short for snapshot?

maybe a more descriptive variable name would be useful, like eventListSnapshot

planet/src/main/webapp/js/itinerary.js
```diff
@@ -21,47 +49,133 @@ function renderStartingLocation() {
...
+    const order = snap.numChildren() + 1;
+
+    //Create a new event
+    const newEventRef = eventListRef.push();
+    newEventRef.set({
```
EkkiD: does this persist the new event in the database?

EkkiD (changes requested):

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
zhoua0805: Not sure why it looks off in github, since it looks fine in my local version...

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
zhoua0805: Okay!

zhoua0805: It would be an internal server error I think. I just updated the status code.

planet/src/main/webapp/js/itinerary.js
zhoua0805: You're right. Since I used snap quite a lot in this file and in my next PR too, can I address this problem altogether in my next PR?

planet/src/main/webapp/js/itinerary.js
zhoua0805: The push() method generates a unique reference key for the new event, and calling set() will overwrite the values in the new event and save it.

EkkiD: SGTM

planet/src/main/webapp/js/itinerary.js
EkkiD: each PR needs to leave the main branch in a minimum level of healthiness. In this case, this PR should be modified to use descriptive variable names.

EkkiD (changes requested):
After addressing the variable naming issue, this LGTM

planet/src/main/webapp/js/itinerary.js
zhoua0805: Ah okay, fixed!

EkkiD (approved):
LGTM

PR #21, opened 2020-07-10
Save events

zhoua0805:
Changes made:
- Add basic validation for add-event-form
- Create a select tag that deplays the names of the saved lists
- Allow users to save their current list to a new list with their choice of name
  - After the user clicks save, the current list will be emptied
  - Users can then access the saved list by selecting it in the select tag above
  - All new edits in the saved list will be automatically saved in the database

> 
After yesterday's meeting, we decided to make some small changes to the data structure: 

         users --> user1  --> name
                          --> email
                          --> phoneNumber

         events --> user1 --> currentList --> event 1
                                          --> event 2
                          --> savedList1  --> event 1
                                          --> event 2
                          --> savedList2 ...

         places --> user1 --> place1
                          --> place2

Next Steps:
- Start integrating with the places data 
- Allow users to rename their saved lists/ delete their saved lists (NOT REQUIRED FOR MVP)
- Don't allow users to delete their saved list by deleting every event (or give a warning when they attempt to delete the last event in the list) (NOT REQUIRED FOR MVP)

### DIFF
--- a/planet/src/main/webapp/js/itinerary.js
+++ b/planet/src/main/webapp/js/itinerary.js
@@ -15,24 +15,38 @@
 
 import TimeRange from './TimeRange.js';
 // Declare global functions
-window.openForm = openForm;
-window.closeForm = closeForm;
+window.openAddEventForm = openAddEventForm;
+window.closeAddEventForm = closeAddEventForm;
+window.openSaveEventsForm = openSaveEventsForm;
+window.closeSaveEventsForm = closeSaveEventsForm;
 window.handleStartingLocationChange = handleStartingLocationChange;
 window.renderStartingLocation = renderStartingLocation;
+window.handleListOptionChange = handleListOptionChange;
 window.addEvent = addEvent;
+window.saveEvents = saveEvents;
 window.generateItinerary = generateItinerary;
 
 
-function openForm() {
+function openAddEventForm() {
     document.getElementById('add-event').style.display = 'block';
 }
 
 // Clear input values and close the form
-function closeForm() {
+function closeAddEventForm() {
     document.getElementById('add-event-name').value = '';
     document.getElementById('add-event-address').value = '';
     document.getElementById('add-event-duration').value = '';
     document.getElementById('add-event').style.display = 'none';
+}
+
+function openSaveEventsForm() {
+    document.getElementById('save-events-form').style.display = 'block';
+    document.getElementById('save-events-button').style.display = 'none';
+}
+
+function closeSaveEventsForm() {
+    document.getElementById('save-events-button').style.display = 'inline-block';
+    document.getElementById('save-events-form').style.display = 'none';
 }
 
 function handleStartingLocationChange() {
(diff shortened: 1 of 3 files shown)


planet/src/main/webapp/js/itinerary.js
```diff
@@ -49,19 +63,67 @@ function renderStartingLocation() {
     }
 }
 
+function renderListOptions() {
```
EkkiD, line 65: what does 'list' refer to? List of events? List of itineraries?

planet/src/main/webapp/js/itinerary.js
```diff
@@ -49,19 +63,67 @@ function renderStartingLocation() {
...
 async function addEvent() {
     const eventName = document.getElementById('add-event-name').value;
     const eventAddress = document.getElementById('add-event-address').value;
     const eventDuration = document.getElementById('add-event-duration').value;
     const userId = firebase.auth().currentUser.uid; 
+    const listName = document.getElementById('list-options').value;
+
+    // Validate the input fields
+    if ((!eventDuration) || (!eventAddress) || (!eventDuration)) {
+        alert('Please make sure to fill out all fields correctly');
```
EkkiD: We should give the user more specific information about what they messed up. Specifically which fields and what is wrong with that field. If you want, you can split this into a validateEventInput() function to keep this code path clean.

planet/src/main/webapp/js/itinerary.js
```diff
@@ -15,26 +15,40 @@
...
+window.openAddEventForm = openAddEventForm;
+window.closeAddEventForm = closeAddEventForm;
+window.openSaveEventsForm = openSaveEventsForm;
+window.closeSaveEventsForm = closeSaveEventsForm;
 window.handleStartingLocationChange = handleStartingLocationChange;
 window.renderStartingLocation = renderStartingLocation;
+window.handleListOptionChange = handleListOptionChange;
 window.addEvent = addEvent;
+window.saveEvents = saveEvents;
 window.generateItinerary = generateItinerary;
```
emilyzhaoo: LGTM with nit: remove extra line

emilyzhaoo (approved):

planet/src/main/webapp/js/itinerary.js
```diff
@@ -49,19 +63,67 @@ function renderStartingLocation() {
...
+}
+
+function handleListOptionChange() {
+    const listName = document.getElementById ('list-options').value;
+    sessionStorage.setItem('listName', listName);
+    // You can only save currentList as other lists
+    if (listName !== 'currentList') {
+        document.getElementById('save-events-button').style.display = 'none';
+    }else{
+        document.getElementById('save-events-button').style.display = 'inline-block';
```
EkkiD, line 97: How do you save a  list as itself? E.g. I have a list called "test", and I want to save changes to it.

EkkiD (changes requested):

planet/src/main/webapp/js/itinerary.js
zhoua0805: Changes will be automatically saved if users are on other lists (just like how currentList is automatically updated). The save button is for exporting currentList. Hope that makes sense!

planet/src/main/webapp/js/itinerary.js
zhoua0805: It's the list of events. We're not saving itineraries in the back end

planet/src/main/webapp/js/itinerary.js
zhoua0805: Okay!

EkkiD (approved):
LGTM

planet/src/main/webapp/js/itinerary.js
zhoua0805: Fixed!

PR #22, opened 2020-07-10
Set up places column to the itinerary page

zhoua0805:
- Set up places column to the itinerary page
- Render a list of places

Demo:
image

Next steps: 
- enable "add" feature to create an event from the selected place


azhang72 (approved):
Thanks for the screenshot!

nabil-dbz (approved):
The screenshot is a great idea! LGTM

PR #24, opened 2020-07-13
Integrate places with events

zhoua0805:
- Allow users to add an event from places
- Render an input form for event duration at cursor location
- change the add button icon (and enable/disable it) when a place is added as an event/ an event is deleted

Quick Demo: 
image


planet/src/main/webapp/css/itinerary.css
```diff
@@ -47,6 +47,14 @@ h1,h2,h3,h4,h5,h6{
     z-index: 999;
 }
 
+#add-place-to-event-form {
+    position: absolute;
+    display: none;
+    background: white;
+    border: solid 1px;
```
emilyzhaoo, line 54: Instead of px to make the website respondent, try using vh or vw

emilyzhaoo (approved):
LGTM with nit

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -46,3 +59,71 @@ function createPlaceElement(ref, name, address) {
...
+        openingTime: placeDetails.openingTime,
+        closingTime: placeDetails.closingTime,
+        order: order,
+    });
+    closeAddPlaceForm();
+    renderPlaces();
+}
+
+function validatePlaceDuration(duration) {
+    let isValid = true;
```
EkkiD: it looks like this variable isn't needed. Anytime it's set to false, we immediately return. We can just directly return false or true instead.

EkkiD (changes requested):

azhang72 (approved):
lgtm pending EkkiD's comments

planet/src/main/webapp/css/itinerary.css
zhoua0805: Okay!

planet/src/main/webapp/js/placesItinerary.js

EkkiD (approved):
LGTM

PR #25, opened 2020-07-15
update visitors list

zhoua0805:
- Add the user id to the visitors list when a user selects a place
- Delete the user id from the visitors list when the deleted event was previously created from a place

### DIFF
--- a/planet/src/main/webapp/js/itinerary.js
+++ b/planet/src/main/webapp/js/itinerary.js
@@ -235,6 +235,15 @@ function deleteEvent(listName, ref) {
         });
     });
 
+    // Check if this is an event created from a place, and if so, 
+    // remove the user id from the visitors list 
+    const placeRef = database.ref('places/' + ref);
+    placeRef.once('value', (placeSnapshot) => {
+        if (placeSnapshot.val()) {
+            placeRef.child('visitors').child(userId).remove();
+        }
+    });
+
     // Render places again since the icons might need to change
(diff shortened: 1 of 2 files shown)


planet/src/main/webapp/html/itinerary.html
```diff
@@ -47,6 +47,12 @@
         <div id="content-places"> 
             <p> Start planing your itinerary by choosing your liked places from below: </p>
             <div id="places"></div>
+            <div id="add-place-to-event-form">
+                <p>Duration (in hours): 
+                    <input type="number" id="add-place-duration" name="add-place-duration" max="9" min="0" step = "0.5"> </input></p>
```
emilyzhaoo: I see you've set the max hours as 9, but it may be possible that tourists want to spend longer so maybe it would be safer to set it to 12 hours?

emilyzhaoo (approved):

planet/src/main/webapp/html/itinerary.html
zhoua0805: The itinerary generator generates an itinerary within the shortest opening hours. I was expecting it shorter than 8am - 5pm, which is 9 hours. If the user selects an event that has a duration of 10 hours, the returned itinerary will be empty. I think the chance of tourists visiting a place for 12 hours is unlikely though. 
What are your thoughts?

EkkiD (approved):
LGTM

PR #26, opened 2020-07-15
Add saved place to database with place details

emilyzhaoo:
Move places.js and places.html out of respective folders to resolve errors.
Update firebase database with new saved place and retrieved place details.
Place ID still needs to be retrieved (will think of a solution).

### DIFF
--- a/planet/src/main/webapp/places.js
+++ b/planet/src/main/webapp/places.js
@@ -152,7 +152,7 @@ function addPlaceDetails(places) {
         let placeId = place['place_id']; 
         let url = 'https://maps.googleapis.com/maps/api/place/details/json?placeid=' + placeId + 
         '&key=AIzaSyDK36gDoYgOj4AlbCqh1IuaUuTlcpKF0ns&fields=photo,formatted_phone_number,formatted_address,opening_hours,website';        
-        let placeDetails = {Name: '', Rating: '', Address: '', Photo: '',Phone: '',Hours: '',Website: ''}; 
+        let placeDetails = {Name: '', Rating: '', Address: '', Photo: '',Phone: '',Hours: '',Website: '',Opening: '',Closing: ''}; 
         
         // Fetch url and add photo reference to a list
         promises = [];
@@ -173,10 +173,19 @@ function addPlaceDetails(places) {
             }
             if (response.result.opening_hours != undefined){
                 placeDetails['Hours'] = response.result.opening_hours.weekday_text;
+                // Open and closing time for Monday (MVP purposes)
+                if (response.result.opening_hours.periods[1] != undefined) {
+                    placeDetails['Opening'] = response.result.opening_hours.periods[1].open.time;
+                    if (response.result.opening_hours.periods[1].close != undefined) {
+                        placeDetails['Closing'] = response.result.opening_hours.periods[1].close.time;
+                    }
+                }
             }
             if (response.result.website != undefined) {
                 placeDetails['Website'] = response.result.website;
             }
+
+           
             placeInfo.push(placeDetails);
         }));
     });
@@ -254,6 +263,8 @@ function listResults() {
         let p2 = document.createElement('p');
         let p3 = document.createElement('p');
         let p4 = document.createElement('p');
+        let p5 = document.createElement('p');
+        let p6 = document.createElement('p');
         let a = document.createElement('a');
 
         // Set variables for place details
@@ -262,6 +273,8 @@ function listResults() {
         let address = document.createTextNode(placeInfo[i]['Address']);
         let phoneNumber = document.createTextNode('Phone Number: ' + placeInfo[i]['Phone']);
         let openingHours = document.createTextNode('Opening Hours: ' + placeInfo[i]['Hours']);
+        let open = document.createTextNode(placeInfo[i]['Opening']);
+        let close = document.createTextNode(placeInfo[i]['Closing']);
         let img = document.createElement('img');
         
         // Create and add save icon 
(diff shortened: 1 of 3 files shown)


planet/src/main/webapp/places.js
```diff
@@ -315,9 +342,32 @@ function savePlace(x) {
...
+/** Update database and add place with information */
+function updateDatabase(placeID, name, address, open, close) {
+    database.ref('places/' + placeID).set({
+        name: name,
+        address: address,
+        openingTime: open,
+        closingTime: close
+    });
+}
+
```
zhoua0805, line 369: Do we want to write to the update the users tree as well?

planet/src/main/webapp/places.js
```diff
@@ -315,9 +342,32 @@ function savePlace(x) {
     if(x.innerHTML === "favorite_border") {
         // Set as saved
         x.innerHTML = "favorite";
+        // Save place to database
+        let name = document.getElementById('place-name').innerHTML;
```
zhoua0805, line 346: Wouldn't every card have the same id 'place-name', 'place-address', or 'openingTime'? How do we make sure we're getting the right information here?

planet/src/main/webapp/places.js
emilyzhaoo: Yes! I will do that as a next step, still trying to figure out how to access place ID!

planet/src/main/webapp/places.js
emilyzhaoo: The onclick function is passing through "this" as a parameter meaning the current card!

planet/src/main/webapp/places.js
```diff
@@ -173,10 +173,19 @@ function addPlaceDetails(places) {
...
+                    if (response.result.opening_hours.periods[1].close != undefined) {
+                        placeDetails['Closing'] = response.result.opening_hours.periods[1].close.time;
+                    }
+                }
             }
             if (response.result.website != undefined) {
                 placeDetails['Website'] = response.result.website;
             }
+
+
```
EkkiD, line 188: nit: is this whitespace necessary? Should it just be one line?

planet/src/main/webapp/places.js
```diff
@@ -315,9 +342,32 @@ function savePlace(x) {
     if(x.innerHTML === "favorite_border") {
         // Set as saved
         x.innerHTML = "favorite";
+        // Save place to database
+        let name = document.getElementById('place-name').innerHTML; 
+        let address = document.getElementById('place-address').innerHTML; 
+        let open = document.getElementById('openingTime').innerHTML;
+        let close = document.getElementById('closingTime').innerHTML;
+        updateDatabase('test', name, address, open, close);
```
EkkiD, line 350: Looks like there's still some debugging code here with the hardcoded 'test' string.

planet/src/main/webapp/places.js
```diff
@@ -315,9 +342,32 @@ function savePlace(x) {
...
+    database.ref('places/' + placeID).set({
+        name: name,
+        address: address,
+        openingTime: open,
+        closingTime: close
+    });
+}
+
+/** Delete place from database when unsaved by user */
+function deletePlace(placeID) {
```
EkkiD, line 371: if two users star a place, and then one unstars it, what should happen?

EkkiD (changes requested):

planet/src/main/webapp/places.js
zhoua0805: Ah I see. x is the current card button being clicked right? x is not used in  `document.getElementById()` (line346) though. Ignore this comment if you already tested this with multiple cards and it worked.

planet/src/main/webapp/places.js
emilyzhaoo: Good point, the saved places should be separate and specific to each user's profile so I'll figure out how to implement that! I'm going to remove the "deletion portion" of the code and follow up with that in a separate PR

planet/src/main/webapp/places.js
emilyzhaoo: Yes! I'm still working on how to access the PlaceID from just a card with details displayed on it. This will be updated in a follow up PR

planet/src/main/webapp/places.js
EkkiD: +1 I share the concern about how this behaves with multiple cards.

In addition, it looks like we're misusing the html id property here. We probably want to use a class instead of id:

https://developer.mozilla.org/en-US/docs/Web/API/Element/id

"If the id value is not the empty string, it must be unique in a document."

EkkiD (changes requested):

planet/src/main/webapp/places.js
emilyzhaoo: ^^ I've tested the document.getElementByID() with multiple cards and it does return the correct value. I do understand the concern that all the id's are the same. I've tried switching it to use class-name but when trying to access document.getElementsByClass.innerHTML -> it returns undefined cause it seems that you can't only access the current element with the class name

EkkiD: when getting elements by classname, you get an array back. If there's one element in the list that you want to access, you can get it by indexing the array: parentDOM.getElementsByClassName("test")[0];

EkkiD (approved):
If we've confirmed that this code behaves correctly with multiple places, then we don't need to block the merge based on the use of ID vs class. We can follow up if we discover a bug later

LGTM

zhoua0805 (approved):
LGTM

PR #27, opened 2020-07-16
Generate Usernames

nabil-dbz:
This pull request adds the feature of generating unique usernames for the users when they first sign up.


planet/src/main/webapp/js/authentication.js
```diff
@@ -67,27 +66,52 @@ function resetForm(elementsClass) {
     }
 }
 
-function signUp() {
+async function generateUsername(displayName) {
+    const names = displayName.split(' ');
+    const existingUsernames = [];
+    let username = names[0].toLowerCase() + '.' + names[1].toLowerCase();
```
zhoua0805, line 72: Just curious: do usernames usually contain dot? if people can access profiles through the username, I feel like https://example.com/profile/johnsmith is more common than https://example.com/profile/john.smith?
LGTM otherwise

zhoua0805 (approved):

azhang72 (approved):

planet/src/main/webapp/js/authentication.js
nabil-dbz: Some of them do some of them don't. LinkedIn puts dashes and Facebook uses dots!

PR #29, opened 2020-07-19
Profile page

nabil-dbz:
This pull request allows the user to access their profile page and the profile page of other users through their usernames by accessing the URL: "/user/username". If the username exists in the database, the profile gets displayed. Otherwise, "Profile Not Found" is displayed instead.
TO DO's:
- Allow the users to change information about themselves on their profile (location, bio, display name)
- Add different sections to the profile page (About, Events, Posts)
- Allow the user to add a profile picture


planet/src/main/java/com/google/planet/servlets/UserServlet.java
```diff
@@ -0,0 +1,64 @@
...
+      public void onDataChange(DataSnapshot dataSnapshot) {
+        if (!dataSnapshot.exists()) {
+          System.out.println("Profile Not Found");
+          return;
+        }
+      }
+      
+      @Override
+      public void onCancelled(DatabaseError databaseError) {
+        // This method has to be overrided that's why I'm adding it
```
azhang72, line 43: How/if this method gets triggered? When it does, what will the user sees?

planet/src/main/java/com/google/planet/servlets/UserServlet.java
```diff
@@ -0,0 +1,64 @@
...
+    UserServlet.initializeFirebase();
+    final FirebaseDatabase database = FirebaseDatabase.getInstance();
+    final DatabaseReference reference = database.getReference("/users");
+    reference.orderByChild("username").equalTo(username).addListenerForSingleValueEvent(
+    new ValueEventListener() {
+      @Override
+      public void onDataChange(DataSnapshot dataSnapshot) {
+        if (!dataSnapshot.exists()) {
+          System.out.println("Profile Not Found");
+          return;
```
azhang72: nit: return statement is redundant.

planet/src/main/webapp/js/profile.js
```diff
@@ -0,0 +1,143 @@
...
+        case 'posts-section':
+            document.getElementById('about-section').style.display = 'none';
+            document.getElementById('posts-section').style.display = 'block';
+            document.getElementById('events-section').style.display = 'none';
+            break;
+        case 'events-section':
+            document.getElementById('about-section').style.display = 'none';
+            document.getElementById('posts-section').style.display = 'none';
+            document.getElementById('events-section').style.display = 'block';
+            break;
```
azhang72, line 31: Add a default block for some error reporting (e.g. "ERROR: unhandled section ID " + sectionId)

planet/src/main/webapp/js/profile.js
```diff
@@ -0,0 +1,143 @@
...
+            document.getElementById('about-link').classList.add('active');
+            document.getElementById('posts-link').classList.remove('active');
+            document.getElementById('events-link').classList.remove('active');
+            break;
+        case 'posts-link':
+            document.getElementById('about-link').classList.remove('active');
+            document.getElementById('posts-link').classList.add('active');
+            document.getElementById('events-link').classList.remove('active');
+            break;
+        case 'events-link':
```
azhang72, line 49: Add a default block

azhang72 (changes requested):

planet/src/main/java/com/google/planet/servlets/UserServlet.java
nabil-dbz: "This method will be triggered in the event that this listener either failed at the server, or is removed as a result of the security and Firebase Database rules." I will print something when this method gets triggered!

planet/src/main/java/com/google/planet/servlets/UserServlet.java
nabil-dbz: Fixed

planet/src/main/webapp/js/profile.js
nabil-dbz: Fixed

planet/src/main/webapp/js/profile.js
nabil-dbz: Fixed

zhoua0805 (approved):
LGTM

planet/src/main/webapp/js/profile.js
```diff
@@ -0,0 +1,147 @@
...
+        return;
+    }
+    document.getElementById('not-found-message').remove();
+    document.getElementById('profile-page').style.display = 'block';
+    let user;
+    let userId;
+    for (const object in userSnapshot.val()) {
+        userId = object;
+        user = userSnapshot.val()[object];
+    }
```
azhang72: What is the point of this for loop? You just keep overwriting |userId| and |user| until you get the last one?

nabil-dbz: The userSnapshot.val() returns an object of users. The user I want to get is only accessed through the id (which I don't have) that's why I have to loop through the properties of the object (which are user ids). Since usernames are unique, the returned object will always contain one user.

azhang72: Can you retrieve the id with something like
const userId = userSanpshot.val()[0] ? (this is where my JS knowledge reaches its limit :P)

nabil-dbz: No! Actually that's not possible because it's an object not an array.
P.S: You can access the property "prop" of the object using: object["prop"] or object.prop, but you need the name of the property which in this case is the user id.

azhang72: Ah I see. How about rename |object| to |property| and leave a comment mentioning that userSnapShot.val() only has one property |username| is unique (also consider renaming |username| to |userId|)?

azhang72 (approved):

PR #34, opened 2020-07-21
Add Location and Bio

nabil-dbz:
This pull request allows the user to add a new location and a new bio on their profile.
addBio
addLocation

### DIFF
--- a/planet/src/main/webapp/js/headerRenderer.js
+++ b/planet/src/main/webapp/js/headerRenderer.js
@@ -1,0 +1,224 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export const HeaderRenderer = {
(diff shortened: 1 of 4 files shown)


planet/src/main/webapp/js/headerRenderer.js
```diff
@@ -0,0 +1,224 @@
...
+        return;
+    }
+    cancelLocationEditing();
+}
+
+function cancelLocationEditing() {
+    renderLocation(undefined, currentUser.uid);
+}
+
+function addLocationToDatabse(location) {
```
EkkiD: typo: Database

planet/src/main/webapp/js/headerRenderer.js
```diff
@@ -0,0 +1,224 @@
...
+    const userReference = database.ref('users/' + currentUser.uid);
+    userReference.update({
+        location: location
+    }).then(() => {
+        renderLocation(location, currentUser.uid);
+    });
+}
+
+function addNewBio() {
```
EkkiD: Generally, functions should be named using the verb that the function is doing, not the verb that the user is doing.

In this case, it looks like the function is opening the Add New Bio dialog. Maybe it could be called openAddNewBioDialog or openBioDialog?

EkkiD (changes requested):

planet/src/main/webapp/js/headerRenderer.js
nabil-dbz: Fixed!

planet/src/main/webapp/js/headerRenderer.js
nabil-dbz: I changed it to "addNewBioForm".

zhoua0805 (approved):
LGTM

EkkiD (approved):
LGTM

PR #35, opened 2020-07-22
get place details from the google maps place details api

zhoua0805:
- Modify renderPlaces() and submitPlace() by getting place details from the google maps api instead of from the database
- Modified version ensures that the place details are always up to date


planet/src/main/java/com/google/planet/data/Itinerary.java
```diff
@@ -0,0 +1,28 @@
...
+
+public final class Itinerary {
+    public final List<ItineraryItem> itineraryItems; 
+    public final String errorMessage;
+
+    public Itinerary(List<ItineraryItem> itineraryItems, String errorMessage) {
+        this.itineraryItems = itineraryItems;
+        this.errorMessage = errorMessage;
+    }
+}
```
nabil-dbz: nit: empty line

emilyzhaoo (approved):
LGTM

planet/src/main/java/com/google/planet/data/TimeRange.java
```diff
@@ -15,9 +15,9 @@
 package com.google.planet.data;
 
 public final class TimeRange {
-    public static final int START_OF_DAY = getTimeInMinutes(8, 0);
-    public static final int END_OF_DAY = getTimeInMinutes(19, 59);
-    public static final TimeRange WHOLE_DAY = new TimeRange(getTimeInMinutes(8,0), 12 * 60);
+    public static final int START_OF_DAY = getTimeInMinutes(0, 0);
+    public static final int END_OF_DAY = getTimeInMinutes(23, 59);
+    public static final TimeRange WHOLE_DAY = new TimeRange(getTimeInMinutes(0,0), 24 * 60);
```
nabil-dbz, line 20: Why have you decided to switch back to 00:00 -> 23:59?

zhoua0805: It was 8 am to 7pm before, but then some restaurants have opening hours till very late (11pm etc.) and the timerange class will return an error because of that

nabil-dbz: I see! okay

nabil-dbz (approved):

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -13,55 +13,104 @@
...
+        }
     });
 }
 
-function createPlaceElement(ref, name, address) {
+function renderPlaceDetailsCallback(place, status) {
+    if (status == google.maps.places.PlacesServiceStatus.OK) {
+        const placesContainer = document.getElementById('places');
+        const placeElement = createPlaceElement(place);
+        placesContainer.appendChild(placeElement);
```
EkkiD, line 70: do these need to be rendered in a particular order?

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -13,55 +13,104 @@
...
     return placeElement;
 }
 
+function createParagraphElement(text){
+    const p = document.createElement('p');
+    p.innerText = text;
+    return p;
+}
+
 function openAddPlaceForm(event, ref) {
```
EkkiD: what type is ref? Consider renaming this variable to something more descriptive. (I think it's a place id?)

EkkiD (changes requested):

planet/src/main/java/com/google/planet/data/TimeRange.java
```diff
@@ -64,8 +64,8 @@ private static boolean equals(TimeRange a, TimeRange b) {
     }
 
     public static int getTimeInMinutes(int hours, int minutes) {
-        if (hours < 8 || hours > 19) {
-            throw new IllegalArgumentException("Hours can only be 8 through 19 (inclusive).");
+        if (hours < 0 || hours > 23) {
```
azhang72, line 67: These changes are not relevant to the PR (get place details from the google maps place details api). You can separate changes like these in another PR. It will land faster and offers more clarity to whoever looks at the commit history in the future.

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -137,3 +159,54 @@ function validatePlaceDuration(duration) {
...
+            // for MVP purposes, use Monday's hours. 
+            const openHours = place.opening_hours.periods[1].open.hours;
+            const openMinutes = place.opening_hours.periods[1].open.minutes;
+            const closeHours = place.opening_hours.periods[1].close.hours;
+            const closeMinutes = place.opening_hours.periods[1].close.minutes;
+            openingTime = TimeRange.getTimeInMinutes(openHours, openMinutes);
+            closingTime = TimeRange.getTimeInMinutes(closeHours, closeMinutes);
+        } else { 
+            // Set the event's opening hours to 8am - 5pm otherwise. 
+            openingTime = TimeRange.getTimeInMinutes(8, 0);
```
azhang72: Is there some way to inform users that these opening hours might not be accurate since the Place API doesn't have them?

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -137,3 +159,54 @@ function validatePlaceDuration(duration) {
     }
     return true;
 }
+
+async function submitPlaceCallback(place, status) {
+    if (status == google.maps.places.PlacesServiceStatus.OK) {
```
azhang72: Consider to have some kind of error message to let users know their submission failed.

planet/src/main/java/com/google/planet/data/Itinerary.java
zhoua0805: Fixed!

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: It doesn't have to be any particular order. I think this function has to modified (in my next PR) because google maps javascript api can't make more than 10 requests at a time. We're not sure the best way to solve it however.
image
https://stackoverflow.com/questions/41946356/google-places-api-batch-requests-javascript 

Option 1 is to have a load more button that displays 10 more cards every time its clicked. The cons are it's hard to keep track of what has been loaded and what hasn't. And since the bucket is refilled 1 second at a time that means users can click "load more" twice in 10 seconds

Option 2 is to go back to saving place details in the database, but somehow update the entire database once in a while (once a month etc.). Not sure how to implement it though

Option 3 is saving the only the name a address in the database (since they likely won't change), and only get place details when the user adds the place as an event. The con is that the user won't be able to see place details (ratings, opening hours...) on the itinerary page, and they could have been useful information for users to decide which places to add and which places not to.

What do you guys think @emilyzhaoo @nabil-dbz @EkkiD @StarIU072

planet/src/main/webapp/js/placesItinerary.js

planet/src/main/java/com/google/planet/data/TimeRange.java
zhoua0805: Okay!

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: Fixed

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: I can just display it on the card :D

planet/src/main/webapp/js/placesItinerary.js
EkkiD: I would focus on Option 1. One idea would be to keep track with a setTimeout() counter that indicates the number of requests you can make. If the user clicks "load more" you just load as many as you've counted. 

That is a hack to work around the challenge, but I think it might be the easiest short-term approach

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -13,56 +13,107 @@
...
     const addButton = document.createElement('div');
     addButton.setAttribute('class', 'place-button');
     
-    const eventFromThisPlace = document.getElementById(ref);
+    const eventFromThisPlace = document.getElementById(place.place_id);
     if (eventFromThisPlace) {
         addButton.innerHTML = `<i class="material-icons small">playlist_add_check</i>`;
     }else {
-        addButton.setAttribute('onclick', 'openAddPlaceForm(event, "' + ref + '")');
+        addButton.setAttribute('onclick', 'openAddPlaceForm(event, "' + place.place_id + '")');
```
EkkiD, line 116: I'm trying to understand this logic. If this is a button on an existing place, you already have the place object. Could you store that object somewhere so that you don't need to load it from the places API when you click thebutton to add it to the itinerary?

One option would be to store it as a map of place id -> place object in the module's namespace
I'm sure there are other options as well.

EkkiD (changes requested):

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: Okay! Can I do it in my next PR?

zhoua0805: The code is currently working such that if the user saves more than 10 places then only 10 cards will show but no errors

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: Do you mean when users are adding a place to the itinerary, we don't have to make another api call, instead we can just get the place information from the dom? (maybe assigning the elements ids like `place-name-placeid`, and `place-address-placeid`)?

EkkiD: > Do you mean when users are adding a place to the itinerary, we don't have to make another api call,

Yes that's what I mean.

> maybe assigning the elements ids like place-name-placeid, and place-address-placeid

That sounds like a good option if you're planning to display all of that data. If not, storing them in a JS object is probably cleaner.

zhoua0805: Sounds good, I made the place objects :)

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -13,56 +13,140 @@
...
+import TimeRange from './TimeRange.js';
+
+// Declare global functions.
 window.openAddPlaceForm = openAddPlaceForm;
 window.closeAddPlaceForm = closeAddPlaceForm;
 
+// Declare global variables/constants.
 const database = firebase.database();
+let map = new google.maps.Map(document.getElementById('empty-map'));
+let savedPlaces;  // Object that contains all the places information
```
EkkiD: nit: init this to empty object here too: {}

EkkiD: also, maybe "displayedPlaces" is a more accurate variable name?

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -13,56 +13,140 @@
...
 }
 
-function createPlaceElement(ref, name, address) {
+function renderPlaceDetailsCallback(place, status) {
+    if (status == google.maps.places.PlacesServiceStatus.OK) {
+        const placesContainer = document.getElementById('places');
+        const placeElement = createPlaceElement(place);
+        placesContainer.appendChild(placeElement);
+    } else {
+        alert( 'Too many requests at the same time! please try again in another 10s. ');
```
EkkiD: If I'm understanding correctly, the user can't directly control how many requests they're making. An alert is useful if the user can take action to fix the problem, but they can't really do that in this case.

Instead, I think we could try to hide or work around the issue so that we never issue too many requests at once. We can do that in a future PR, but in the meantime I'd still recommend removing the alert to avoid confusion.

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -97,18 +181,14 @@ async function submitPlace(ref) {
     const eventListSnapshot = await eventListRef.once('value');
     const order = eventListSnapshot.numChildren() + 1;
 
-    // Create a new event based on the place
-    const placeRef = database.ref('places/' + ref); 
-    const placeSnapshot = await placeRef.once('value');
-    const placeDetails = placeSnapshot.val();
-
-    const newEventRef = eventListRef.child(ref);
+    const place = savedPlaces[placeId];
```
EkkiD: We should probably check for null in this case. I don't think it's possible for it to be null, but let's at least handle the error gracefully if it is (maybe a console log and skipping the rest of the steps?

EkkiD (changes requested):

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: Okay

planet/src/main/webapp/js/placesItinerary.js

planet/src/main/webapp/js/placesItinerary.js

EkkiD (approved):
LGTM

PR #37, opened 2020-07-22
Add autocomplete to Itinerary Generator

emilyzhaoo:
Add autocomplete to input box in Itinerary Generator. 

Allow the users to enter addresses more efficiently. Implement for starting address and event address.


planet/src/main/webapp/places.js
```diff
@@ -292,6 +285,7 @@ function listResults() {
         icon.classList.add('material-icons');
         icon.classList.add('small');
```
nabil-dbz: nit: you can do this in one line: icon.classList.add('material-icons', 'small')

planet/src/main/webapp/places.js
```diff
@@ -342,36 +336,33 @@ function listResults() {
...
     }
-    // Add search keyword to header
+    // Add search keyword to header 
     document.getElementById('greeting').innerHTML = 'Find a place: ' + document.getElementById('pac-input').value;
 }
 
 /** Toggle save icon on click */
 function savePlace(x) {
+    let placeID = $(x).attr('name');
+    let userID = currentUser.uid;
```
nabil-dbz: Are you sure that currentUser is not 'null'? (the user is logged in)

planet/src/main/webapp/itinerary.html
```diff
@@ -87,4 +87,23 @@
...
+            autocomplete = new google.maps.places.Autocomplete(startAddress,options); 
+            autocomplete = new google.maps.places.Autocomplete(eventAddress,options);
+            autocomplete.setFields(['address_component']);
+            autocomplete.addListener('place_changed', fillAddress);
+        }
+
+        function fillAddress() {
+            let place = autocomplete.getPlace();
+        }
+    </script>
```
nabil-dbz: Why did you add these two functions in the html file?

emilyzhaoo: I was calling them from the JS file and I kept getting the error "initAutocomplete does not exist" and I tried a bunch of solutions such as moving the api call in the html file and adding async defer but in the end they didn't work

planet/src/main/webapp/places.js
emilyzhaoo: When the user is logged in it isn't null

emilyzhaoo: We will have to implement something that tells the user they have to be logged in to save places

planet/src/main/webapp/itinerary.html
emilyzhaoo: So this is the solution I went with

planet/src/main/webapp/places.js
nabil-dbz: If the user is not logged in, you will get an error for trying to access the property uid of undefined. I suggest you add a condition to check that (currentUser !== null).

emilyzhaoo: Will do in next PR!

planet/src/main/webapp/itinerary.html
```diff
@@ -87,4 +87,23 @@
...
     </body>
+    <script>
+        function initAutocomplete() {
+            let startAddress = document.getElementById('starting-address');
+            let eventAddress = document.getElementById('add-event-address');
+
+            let options = {
+                types: ['geocode']
+            };
+            autocomplete = new google.maps.places.Autocomplete(startAddress,options);
```
EkkiD: do you need to do any setFields or addListeners on this object?

planet/src/main/webapp/itinerary.html
EkkiD: Which js file did you have them in?

EkkiD (changes requested):
Add some more info to the PR description with more context: https://google.github.io/eng-practices/review/developer/cl-descriptions.html

planet/src/main/webapp/itinerary.html
emilyzhaoo: Itinerary.js

emilyzhaoo: This is the error: 
Screenshot 2020-07-22 at 11 52 33 AM

planet/src/main/webapp/itinerary.html
emilyzhaoo: Yes, in line 100 setFields() is used to specify which data to return, which in this case is the address component. Line 101 adds a listener to update the input box when a new search is performed

planet/src/main/webapp/itinerary.html
EkkiD: itinerary.js is a module. functions in a module are not added to the global namespace by default.

If you add this function to the global namespace in itinerary.js, what is the behaviour? See lines 18-28 of that file for examples

planet/src/main/webapp/itinerary.html
EkkiD: you're creating the autocomplete object, and them immediately replacing it with a different one. If you mean to setsField and addListener on both of them, you'll need to change this code to use a different variable name.

EkkiD (changes requested):

planet/src/main/webapp/js/authentication.js
```diff
@@ -25,7 +25,7 @@ const firebaseConfig = {
     measurementId: 'G-9W48MVBXLE'
 };
 let currentUser;
-let database;
+let database;
```
azhang72: nit: trailing space

planet/src/main/webapp/js/itinerary.js
```diff
@@ -56,7 +56,7 @@ function handleStartingLocationChange() {
         sessionStorage.setItem('start', document.getElementById('starting-address').value);
     } else {
         alert ('Please update your browser'); 
-    }
+    }
```
azhang72: nit: trailing space

planet/src/main/webapp/places.js
```diff
@@ -147,6 +147,7 @@ function setSearchByButton() {
 function addPlaceDetails() {
     // For each place, get the icon, name and location.
     places.forEach(function(place) {
+
```
azhang72: nit: wandering new line

planet/src/main/webapp/places.js
```diff
@@ -206,7 +200,6 @@ function callback(place, status) {
         }
         placeInfo.push(placeDetails);
     }    
-
```
azhang72: nit: remove unrelated changes

planet/src/main/webapp/places.js
```diff
@@ -342,36 +336,33 @@ function listResults() {
         div1.appendChild(div2); 
         element.appendChild(div1);
     }
-    // Add search keyword to header
+    // Add search keyword to header
```
azhang72: nit: trailing space

azhang72 (commented):
a few nits

planet/src/main/webapp/itinerary.html
zhoua0805: I got the same error when I was implementing the get place details function (PR #35 ), even after I declared it as a global namespace. So I chose to load the google maps js file synchronously, and call initMap at body onload, which solves the problem

emilyzhaoo: Okay! I implemented this in the new commit! thanks

zhoua0805 (approved):
LGTM

planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
```diff
@@ -48,10 +48,12 @@ public Itinerary generateItinerary(List<Event> events) {
 
         Collections.sort(events, Event.SortByOrder);
         Duration [][] travelTimeGraph = getEventsDistanceMatrix(events);
+
```
EkkiD: remove these newlines to keep the diff summary concise

EkkiD (changes requested):

EkkiD (approved):
LGTM

planet/src/main/webapp/js/authentication.js
```diff
@@ -25,7 +25,11 @@ const firebaseConfig = {
     measurementId: 'G-9W48MVBXLE'
 };
 let currentUser;
+<<<<<<< HEAD
+let database;
+=======
```
EkkiD: oops there's still a merge confict here

planet/src/main/webapp/js/itinerary.js
```diff
@@ -26,9 +26,23 @@ window.handleListOptionChange = handleListOptionChange;
...
+/** Adds autocomplete to input boxes */
+function initAutocomplete() {
+    let startAddress = document.getElementById('starting-address');
+    let eventAddress = document.getElementById('add-event-address');
+
+    let options = {
+        types: ['geocode']
+    };
+
+    new google.maps.places.Autocomplete(startAddress,options);
```
EkkiD: Consider assigning these to variables or globals so that they can be referenced in the future if needed

EkkiD (changes requested):

planet/src/main/webapp/js/authentication.js
emilyzhaoo: Oops! my bad

planet/src/main/webapp/js/itinerary.js
```diff
@@ -26,9 +26,23 @@ window.handleListOptionChange = handleListOptionChange;
...
+/** Adds autocomplete to input boxes */
+function initAutocomplete() {
+    let startAddress = document.getElementById('starting-address');
+    let eventAddress = document.getElementById('add-event-address');
+
+    let options = {
+        types: ['geocode']
+    };
+
+    let autocompleteStart = new google.maps.places.Autocomplete(startAddress,options);
```
EkkiD: The code here is still confusing, but I guess it's a quirk of the places API.

We're creating an object with `new`, then binding it to a variable. Then we're immediately letting it fall out of scope, removing any reference to the autocomplete object. In theory, the javascript's garbage collector should then delete the object, which would cause autocomplete to stop working. However, I guess that isn't happening, since you've tested this and it works.

More about garbage collection here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management

I think the most readable solution would be to define variables on the module scope (by declaring them with `let` outside of any function).
Alternatively, add a comment saying that this is safe because the object doesn't get cleaned up by the garbage collector.

EkkiD (changes requested):

planet/src/main/webapp/js/itinerary.js
emilyzhaoo: Okay! Makes sense

planet/src/main/webapp/js/itinerary.js
```diff
@@ -26,9 +26,26 @@ window.handleListOptionChange = handleListOptionChange;
 window.addEvent = addEvent;
 window.saveEvents = saveEvents;
 window.generateItinerary = generateItinerary;
+window.initAutocomplete = initAutocomplete; 
 
 const database = firebase.database();
 
+// Declare global variables 
+// This is safe because the autocomplete objects do not get cleaned up by the garbage collector
```
EkkiD: by making this global, we're explicitly preventing them from getting cleaned up. This comment is not necessary.

EkkiD (changes requested):

planet/src/main/webapp/js/itinerary.js
emilyzhaoo: I see! Okay will remove it

EkkiD (approved):
LGTM

PR #41, opened 2020-07-23
Set up basic firebase security rules

zhoua0805:
- Add a file that contains all the firebase security rules
  - using txt now since I can't add comments to json files
- Give read access of the entire database to all users
- Only give write access to users' own data

### DIFF
--- a/planet/src/main/FirebaseSecurityRules.txt
+++ b/planet/src/main/FirebaseSecurityRules.txt
@@ -18,8 +18,8 @@
         // Rules applied to the users tree
         // Users can read other users' data and write their own data
         "users": {
+            ".read": true,
             "$uid": {
-                ".read": true,
                 ".write": "$uid === auth.uid"
             }
         },
--- a/planet/src/main/FirebaseSecurityRules.txt
+++ b/planet/src/main/FirebaseSecurityRules.txt
@@ -14,32 +14,39 @@
 
 // This file contains the security rules of the firebase realtime databse
 {
-  "rules": {
-    // Rules applied to the users tree
-    "users": {
-      "$uid": {
-        ".read": true,
-        ".write": "$uid === auth.uid"
-      }
-    },
-    
-    // Rules applied to the places tree
-    "places": {
-      "$placeId":{
-        ".read": "root.child('users').child(auth.uid).child('places').child('$placeId').exists() === true",
-        ".write": "root.child('users').child(auth.uid).child('places').child('$placeId').exists() === true"
-      }
-    },
-    
-    // Rules applied to the events tree
-    "events": {
-      "$uid": {
-        ".read": true,
-        ".write": "$uid === auth.uid",
-        "$listName": {
-          ".indexOn": ["order"]
+    "rules": {
+        // Rules applied to the users tree
+        // Users can read other users' data and write their own data
+        "users": {
+            "$uid": {
+                ".read": true,
+                ".write": "$uid === auth.uid"
+            }
+        },
+        
+        // Rules applied to the places tree
+        // Users can read all places data, but can only edit their own data
+        "places": {
+          	".read": true,
+            "$placeId":{
+                "visitors":{
+                    "$uid" : {
+                        ".write": "$uid === auth.uid"
+                    }
+                }
+            }
+        },
+        
+        // Rules applied to the events tree
+        // Users can read other users' data and write their own data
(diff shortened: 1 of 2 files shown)


planet/src/main/FirebaseSecurityRules.txt
```diff
@@ -0,0 +1,52 @@
...
+        },
+        
+        // Rules applied to the events tree
+        // Users can read other users' data and write their own data
+        "events": {
+            "$uid": {
+                ".read": true,
+                ".write": "$uid === auth.uid",
+                "$listName": {
+                    ".indexOn": ["order"]
```
EkkiD: why do we need the index on order?

EkkiD (changes requested):

emilyzhaoo (approved):
LGTM

planet/src/main/FirebaseSecurityRules.txt
```diff
@@ -0,0 +1,52 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file contains the security rules of the firebase realtime databse
+{
+    "rules": {
+        // Rules applied to the users tree
+        // Users can read other users' data and write their own data
+        "users": {
+            "$uid": {
```
nabil-dbz, line 22: In order to get the profile using the username, we need to read the users tree. You haven't set any rules for this tree. I was wondering if it's possible to give access to the admin only? (like us developers)

nabil-dbz: Actually, I think we can just leave it as it is but we'll have to create another tree for usernames that everybody can read from later when we add the privacy feature. For now, since all users can read the entire database data, I think we can just add after the 20th line: ".read": true,

planet/src/main/FirebaseSecurityRules.txt
zhoua0805: Firebase provides indexing such that it can improve query performance when we do sorting/filtering. 
This website talks a little about it: https://howtofirebase.com/firebase-security-rules-88d94606ce4a 
image

planet/src/main/FirebaseSecurityRules.txt
```diff
@@ -0,0 +1,52 @@
...
+                ".read": true,
+                ".write": "$uid === auth.uid"
+            }
+        },
+        
+        // Rules applied to the places tree
+        // Users can read all places data, but can only edit their own data
+        "places": {
+          	".read": true,
+            "$placeId":{
```
nabil-dbz, line 31: Same for the placeId tree. We need to add ".read: true," to be able to read all the list of visitors at once.

planet/src/main/FirebaseSecurityRules.txt
zhoua0805: Okay!

planet/src/main/FirebaseSecurityRules.txt
zhoua0805: It's already true from its parent node :)

nabil-dbz: Okay!

nabil-dbz (approved):
LGTM

planet/src/main/FirebaseSecurityRules.txt
EkkiD: in what cases do we search for an even with a specific order?

EkkiD (changes requested):

planet/src/main/FirebaseSecurityRules.txt
zhoua0805: Every time when we're rendering a list of events, we need to order them by the "order" property. (The order property gets changed every time users reorder the cards on the font end)

EkkiD: An index will make a query for the entry with a specific value of the indexed child very fast. It will not change the results at all if you're fetching all of the items.

It sounds from your description like you're loading all events anyways, and then sorting them. An index won't help with that.

EkkiD (changes requested):

planet/src/main/FirebaseSecurityRules.txt
zhoua0805: Ahh okayy, I'll remove it then

EkkiD (approved):
LGTM

PR #42, opened 2020-07-23
Add Events Section

nabil-dbz:
This pull request adds the list of saved lists of events to the database. It also lists the place's list of visitors. 
events


planet/src/main/webapp/js/authentication.js
```diff
@@ -111,7 +111,7 @@ async function signUp() {
     firebase.auth().createUserWithEmailAndPassword(email, password)
     .then(function() {
         const user = firebase.auth().currentUser;
-        addUserToDatabase(user.uid, displayName, email, phoneNumber, username);
+        addUserToDatabase(user.uid, displayName, phoneNumber, email, username);
```
azhang72: changes like these are not exactly relevant to the PR's title. I understand that it feels natural to group small changes into a PR with other changes but please keep them separate.
Having small, bug specific PRs makes reviewing easier and faster.

planet/src/main/webapp/js/profile-events-renderer.js
```diff
@@ -0,0 +1,257 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+let userId;
+let listIndex = 0;
+let eventsList = [];
+let listName = 'List X';
+
+export const ProfileEventsRenderer = { 
+    init: async (uid) => {
```
azhang72: The init pattern is generally frowned upon since there's no guarantee that init() will be called.
Refactor this into an actual class that takes a user id (along with other required fields) in its constructor.

azhang72 (changes requested):

planet/src/main/webapp/css/profile.css
```diff
@@ -1,5 +1,5 @@
 .profile-container {
-    width: min(65%, 600px);
+    width: min(70%, 800px);
```
emilyzhaoo, line 2: We might want to consider using vh, vw or em instead of px to make this responsive!

planet/src/main/webapp/js/profile-events-renderer.js
```diff
@@ -0,0 +1,257 @@
...
+            listName = listId;
+            return getListFromObject(snapshot.val()[listId]);
+        }
+        counter++;
+    }
+    return [];
+}
+
+function getListFromObject(object) {
+    let list = [];
```
emilyzhaoo, line 69: nit: have a more descriptive variable name for this list

emilyzhaoo (approved):
LGTM with nits

planet/src/main/webapp/js/authentication.js
nabil-dbz: Okay! I will add this fix in another PR.

planet/src/main/webapp/css/profile.css
nabil-dbz: It is, actually, responsive since I am using percentages. As for the pixels, they just represent a maximum limit of the width. In other words, If the screen's width is larger than 1143 px, the width of the container shouldn't be more than 800px! But if it's smaller, then, it should take 70% of the width.

planet/src/main/webapp/js/profile-events-renderer.js
nabil-dbz: Okay!

planet/src/main/webapp/js/profile-events-renderer.js
nabil-dbz: Actually, this is just an object with a function that I called init to render the initial list of events. The function gets called in the profile.js script. I think I should probably change its name to something less confusing. 
I don't really need a class but just a module and the reason I am putting the function inside the object is because it allows me to import everything i want to import from this module at once (functions and/or variables).

azhang72 (approved):

PR #43, opened 2020-07-23
Change add place button dynamically

zhoua0805:
This PR fixes some small issues of the itinerary page

**1.** Create function renderPlaceButtons(). 
  - **Context**: The add button of each place needs to be enabled/disabled after user adds a place/ deletes an event that used to be a place/ changes to a new list to work on. 
- Previously, the renderPlaces function is called every time these events happen, which clears the places container and fetches all the place details from the google maps api again. Therefore, a lot of redundant requests are made.
- **Fix**: The new function only changes the button, and keeps everything else the same
- if the parameter placeId is null, the function will go through every place and update their buttons, else, it will only toggle the provided place's button

**2.** Remove the line that calls the sigup.css file in itinerary.html since it's affecting the page layout

**3.** Update starting address in sessionStorage
 - **Context**: The starting address is stored in sessionStorage and is updated using an input onChange function. However, when we use the google maps autocomplete feature to fill out the address, the onChange function skips it (for example, if we type in "to" and google maps autocompletes to "toronto", the value stored in sessionStorage will still be "to"
- **Fix**: add an event listener to the autocomplete to update sessionStorage.

### DIFF
--- a/planet/src/main/webapp/js/placesItinerary.js
+++ b/planet/src/main/webapp/js/placesItinerary.js
@@ -255,7 +255,6 @@ export function renderPlaceButtons(placeId, enableAdd) {
         placeListRef.once('value', (placeListSnapshot) => {
             placeListSnapshot.forEach(function(childPlace) {
                 let placeId = childPlace.key;
-                console.log(document.getElementById(placeId));
                 const noSuchEvent = document.getElementById(placeId) === null;
                 togglePlaceButton (placeId, noSuchEvent);
             });
--- a/planet/src/main/webapp/js/placesItinerary.js
+++ b/planet/src/main/webapp/js/placesItinerary.js
@@ -231,7 +231,7 @@ async function submitPlace(placeId) {
     });
 
     closeAddPlaceForm();
-    renderPlaceButtons(placeId, false);
+    disablePlaceButton(placeId);
 }
 
 function validatePlaceDuration(duration) {
@@ -246,31 +246,37 @@ function validatePlaceDuration(duration) {
     return true;
 }
 
-export function renderPlaceButtons(placeId, enableAdd) {
-    if (placeId) {
-        togglePlaceButton(placeId, enableAdd);
-    } else {
-        const userId = firebase.auth().currentUser.uid;
-        const placeListRef = database.ref('users/' + userId + '/places');
-        placeListRef.once('value', (placeListSnapshot) => {
-            placeListSnapshot.forEach(function(childPlace) {
-                let placeId = childPlace.key;
-                const noSuchEvent = document.getElementById(placeId) === null;
-                togglePlaceButton (placeId, noSuchEvent);
-            });
-        });
+export function renderPlaceButtons() {
+    if (firebase.auth().currentUser === null) {
+        return;
     }
+    const userId = firebase.auth().currentUser.uid;
+    const placeListRef = database.ref('users/' + userId + '/places');
+    placeListRef.once('value', (placeListSnapshot) => {
+        placeListSnapshot.forEach(function(childPlace) {
+            let placeId = childPlace.key;
+            const eventExists = document.getElementById(placeId) !== null;
+            if (eventExists) {
+                disablePlaceButton(placeId);
+            } else {
+                enablePlaceButton(placeId);
+            }
+        });
+    });
 } 
 
-function togglePlaceButton(placeId, enableAdd) {
+export function enablePlaceButton(placeId) {
     const placeButton = document.getElementById('place-button-' + placeId);
     if (placeButton) {
-        if (enableAdd) {
-            placeButton.setAttribute('onclick', 'openAddPlaceForm(event, "' + placeId + '")');
-            placeButton.innerHTML = `<i class="material-icons small">playlist_add</i>`;
-        } else {
-            placeButton.onclick = null;
-            placeButton.innerHTML = `<i class="material-icons small">playlist_add_check</i>`;
-        }
+        placeButton.setAttribute('onclick', 'openAddPlaceForm(event, "' + placeId + '")');
+        placeButton.innerHTML = `<i class="material-icons small">playlist_add</i>`;
+    }
+}
+
+export function disablePlaceButton(placeId) {
+    const placeButton = document.getElementById('place-button-' + placeId);
+    if (placeButton) {
+        placeButton.onclick = null;
+        placeButton.innerHTML = `<i class="material-icons small">playlist_add_check</i>`;
     }
 }
--- a/planet/src/main/webapp/js/placesItinerary.js
+++ b/planet/src/main/webapp/js/placesItinerary.js
@@ -108,6 +108,7 @@ function createPlaceElement(place) {
     // Create the add button
     const addButton = document.createElement('div');
     addButton.setAttribute('class', 'place-button');
+    addButton.setAttribute('id', 'place-button-' + place.place_id);
     
     const eventFromThisPlace = document.getElementById(place.place_id);
     if (eventFromThisPlace) {
@@ -222,7 +223,7 @@ async function submitPlace(placeId) {
     });
 
     closeAddPlaceForm();
-    renderPlaces();
+    renderPlaceButtons(placeId, false);
 }
 
 function validatePlaceDuration(duration) {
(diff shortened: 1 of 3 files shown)


planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -236,3 +245,32 @@ function validatePlaceDuration(duration) {
     }
     return true;
 }
+
+export function renderPlaceButtons(placeId, enableAdd) {
+    if (placeId) {
+        togglePlaceButton(placeId, enableAdd);
+    } else {
+        const userId = firebase.auth().currentUser.uid;
```
nabil-dbz: It's safer to check that the currentUser is different than null before trying to access the uid property.

zhoua0805: I can quickly add it in, but there won't be any place buttons if the user is not signed in, so it's almost impossible for this edge case to happen

azhang72: Is it possible for the call to firebase.auth().currentUser.uid to fail? Trying to think of any scenarios where |userId| could be null here..

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -236,3 +245,32 @@ function validatePlaceDuration(duration) {
     }
     return true;
 }
+
+export function renderPlaceButtons(placeId, enableAdd) {
```
EkkiD: I don't follow what the second parameter here does. What is the different behaviour if it's true or false?

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -236,3 +245,32 @@ function validatePlaceDuration(duration) {
...
+        });
+    }
+} 
+
+function togglePlaceButton(placeId, enableAdd) {
+    const placeButton = document.getElementById('place-button-' + placeId);
+    if (placeButton) {
+        if (enableAdd) {
+            placeButton.setAttribute('onclick', 'openAddPlaceForm(event, "' + placeId + '")');
+            placeButton.innerHTML = `<i class="material-icons small">playlist_add</i>`;
```
EkkiD: all of the actual logic of this function is different depending on the enableAdd parameter. It seems it it would be clearer as two functions with descriptive names

Reference:
https://docs.google.com/document/d/1R5iFoLp41lbIrHvOpViblqH91j-VscdkXNNAL6uYxlU/edit

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -236,3 +245,32 @@ function validatePlaceDuration(duration) {
...
+    if (placeId) {
+        togglePlaceButton(placeId, enableAdd);
+    } else {
+        const userId = firebase.auth().currentUser.uid;
+        const placeListRef = database.ref('users/' + userId + '/places');
+        placeListRef.once('value', (placeListSnapshot) => {
+            placeListSnapshot.forEach(function(childPlace) {
+                let placeId = childPlace.key;
+                const noSuchEvent = document.getElementById(placeId) === null;
+                togglePlaceButton (placeId, noSuchEvent);
```
EkkiD: use positive booleans: https://docs.google.com/document/d/1ljFkOXo6NROsfKGtASDPRTPUq0oDPpm8jc2D5hBfqLE/preview

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -38,10 +38,18 @@ function initAutocomplete() {
...
+    autocompleteStart = new google.maps.places.Autocomplete(startAddress,options);
+    autocompleteStart.setFields(['address_component']);
+    autocompleteStart.addListener('place_changed', updateStartingAddress);
+
     autocompleteEvent = new google.maps.places.Autocomplete(eventAddress,options);
 }
 
+function updateStartingAddress() {
+    const startAddress = document.getElementById('starting-address');
+    sessionStorage.setItem('start', startAddress.value);
```
EkkiD, line 50: is this related to this PR? It looks like a separate fix

EkkiD (changes requested):

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: I talked about it in the PR description (since all 3 fixes are relatively small I combined them into one PR)

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: If a place haven't been added, the button should be enabled, and it will be disabled as soon as the user adds the place.
Meanwhile, if the user deletes an event that was added from a place, the button should be enabled again.

> Example:

      user has a place called CN Tower --> user adds CN Tower as an event --> button disabled
                                       --> user deletes the CN Tower event --> button enabled

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: hmm not sure, if the user is signed in I can't think of any other scenarios that the userid would be null

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: Ah okay, do you mean I should change the parameter name of enableAdd? If so, what should I change it to?
Or do you mean I should change the function to two functions such as enablePlaceButton() and disablePlaceButton()

planet/src/main/webapp/js/placesItinerary.js

planet/src/main/webapp/js/placesItinerary.js
nabil-dbz: If in you're sure that the function is only called when the user is signed in, then it should be good as it is. I usually take precautions and verify that currentUser is not null to avoid errors popping up in the console.

nabil-dbz (approved):

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: Okay! thanks!

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -38,10 +38,18 @@ function initAutocomplete() {
     let options = {
         types: ['geocode']
     };
-    autocompleteStart = new google.maps.places.Autocomplete(startAddress,options); 
+    autocompleteStart = new google.maps.places.Autocomplete(startAddress,options);
+    autocompleteStart.setFields(['address_component']);
+    autocompleteStart.addListener('place_changed', updateStartingAddress);
+
     autocompleteEvent = new google.maps.places.Autocomplete(eventAddress,options);
```
azhang72, line 45: Will you need similar listeners and fields for |autocompleteEvent| as well?

planet/src/main/webapp/js/placesItinerary.js
azhang72: What about network failure?

zhoua0805: true, i'll add it in then! thank you!

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: When I tested it it seems like the input values are automatically changed with autocomplete. I only need the listener for autocompleteStart since we need to store the start location in sessionStorage

planet/src/main/webapp/js/placesItinerary.js
EkkiD: > Or do you mean I should change the function to two functions such as enablePlaceButton() and disablePlaceButton()

This is what I meant

planet/src/main/webapp/js/placesItinerary.js
EkkiD: It looks like we're putting several different functions into one here and controlling the code path with parameters. That makes it very difficult to read.

Q: What does renderPlaceButtons do?
A: Well, it renders all the places that the user has saved, except when it's called with a place id, in which case it only renders that place. It accepts an enableAdd boolean, but it only uses it if placeId is also set.

A function should do exactly one thing, which should be described by its name. Specifically to this case, this should probably be 3 functions:
disablePlaceButton(placeId)
enablePlaceButton(placeId)
renderPlaceButtons()

EkkiD (changes requested):

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: That makes sense! Thank you :D

EkkiD (approved):
LGTM

PR #44, opened 2020-07-24
Display all 20 place detail cards

emilyzhaoo:
Display all 20 place detail cards.

Work around the 10 place request query limit. Make requests in groups of 10 by splitting up the list of places into 2 and using setTimeout() to call the second list 5 seconds later. Gives Place Request Query time to empty before making more requests.

### DIFF
--- a/planet/src/main/webapp/places.js
+++ b/planet/src/main/webapp/places.js
@@ -162,6 +162,9 @@ function addPlaceDetails() {
     const firstHalf = places.splice(0,half);
     const secondHalf = places.splice(-half);
 
+    // Clear results 
+    document.getElementById('results').innerHTML = ""; 
+
     // For each place, get the icon, name and location.
     firstHalf.forEach(function(place) {
         sendPlaceRequest(place);
@@ -267,9 +270,6 @@ function createMarkers(place) {
 /** List results of nearby search on page*/
 function listResults() {
     let element = document.getElementById('results');
-    // Clear results 
-    element.innerHTML = ""; 
-
     // Displays search results on page
     for (let i = 0; i < placeInfo.length; i++) {
         // Create materialize card for each result by creating dom element in HTML
--- a/planet/src/main/webapp/places.js
+++ b/planet/src/main/webapp/places.js
@@ -105,10 +105,23 @@ function updateSearch() {
         placeInfo = [];
 
         bounds = new google.maps.LatLngBounds();
+    
+        // Display markers on map
+        places.forEach(function(place) {
+            if (!place.geometry) {
+                console.log("Returned place contains no geometry");
+                return;
+            }
+            createMarkers(place); 
+            if (place.geometry.viewport) {
+                bounds.union(place.geometry.viewport);
+            } else {
+                bounds.extend(place.geometry.location);
+            }
+        });
         addPlaceDetails(places);
         map.fitBounds(bounds);
     });
-
     // Add onclick function to option buttons
     addOnclick(); 
 }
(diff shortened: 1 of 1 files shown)


planet/src/main/webapp/places.js
```diff
@@ -145,29 +158,39 @@ function setSearchByButton() {
...
-        }
-        createMarkers(place); 
-        if (place.geometry.viewport) {
-            bounds.union(place.geometry.viewport);
-        } else {
-            bounds.extend(place.geometry.location);
-        }
+    const half = Math.ceil(places.length /2); 
+    const firstHalf = places.splice(0,half);
+    const secondHalf = places.splice(-half);
```
EkkiD: if the number of results returned changes (e.g. if the maps API changes its logic), this will break.

We should write code such that it's resilient to changes that we don't control.

planet/src/main/webapp/places.js
```diff
@@ -145,29 +158,39 @@ function setSearchByButton() {
...
-        };
-        // Call Places Details Request
-        let service = new google.maps.places.PlacesService(map);
-        service.getDetails(place, callback); 
+    // For each place, get the icon, name and location.
+    firstHalf.forEach(function(place) {
+        sendPlaceRequest(place);
     });
+
+    // Call remaining 10 place requests after 5 seconds to wait after request query empties
```
EkkiD: The documentation implies that requests are re-issued at a rate of 1 queries per second. Have you confirmed that it regenerates at 2 per second?

planet/src/main/webapp/places.js
```diff
@@ -198,10 +221,10 @@ function callback(place, status) {
...
         }
         placeInfo.push(placeDetails);
-    }    
+    }
     // List results when all callbacks are finished
     count ++ 
-    if (count === places.length) {
+    if (count === 10) {
         count = 0; 
         listResults();
```
EkkiD: doesn't listresults clear the currently displayed results? If so, the second set of requests will replace the first.

I'd suggest taking a step back and thinking holistically about what the sequence of events should be. It might lead you to a cleaner solution to this problem. Happy to help with this design process if that's helpful

EkkiD (changes requested):

planet/src/main/webapp/places.js
emilyzhaoo: yes! I've tested the settimeout function and the fastest time possible to have all the results go through is 5 seconds

planet/src/main/webapp/places.js
```diff
@@ -145,29 +158,39 @@ function setSearchByButton() {
...
     });
+
+    // Call remaining 10 place requests after 5 seconds to wait after request query empties
+    // Place Details Requests can only handle a query of 10 requests
+    setTimeout(function () {
+        secondHalf.forEach(function(place) {
+            sendPlaceRequest(place);
+        })
+    },5000);
+}
```
nabil-dbz: Are we sure it's always 5 seconds? If not, what happens if we can't request all 10 places after 5 seconds?

emilyzhaoo: I've tested it and the requests would complete simultaneously, but in the end they all appear

planet/src/main/webapp/places.js
```diff
@@ -145,29 +158,45 @@ function setSearchByButton() {
...
-        if (place.geometry.viewport) {
-            bounds.union(place.geometry.viewport);
-        } else {
-            bounds.extend(place.geometry.location);
-        }
+    // Clear results 
+    document.getElementById('results').innerHTML = ""; 
+
+    // Place Details Requests can only handle a query of 10 requests
+    let splice = places.splice(0,10);
```
EkkiD: 10 shows up as a magic number a few times in this code.  Prefer adding a CONSTANT above or a local const to indicate what this is.

https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants

planet/src/main/webapp/places.js
```diff
@@ -145,29 +158,45 @@ function setSearchByButton() {
...
-        let service = new google.maps.places.PlacesService(map);
-        service.getDetails(place, callback); 
+    // For each place, get the icon, name and location.
+    splice.forEach(function(place) {
+        sendPlaceRequest(place);
     });
+
+    // Place Details API can call place requests in groups of 20, 40, 60, etc. by using tokens
+    // Allow for all place requests to be called
+    while (places.length > 9) {
```
EkkiD: If there are 19 places, how many will you render with this code?

planet/src/main/webapp/places.js
```diff
@@ -145,29 +158,45 @@ function setSearchByButton() {
...
+        },5000);
+    }
+}
+
+/** Create place requests */
+function sendPlaceRequest(place) {
+    if (!place.geometry) {
+        console.log("Returned place contains no geometry");
+        return;
+    }
```
EkkiD: It looks like geometry is unused here. Do you need this check?

planet/src/main/webapp/places.js
```diff
@@ -198,10 +227,10 @@ function callback(place, status) {
             placeDetails['Website'] = place.url;
         }
         placeInfo.push(placeDetails);
-    }    
+    }
     // List results when all callbacks are finished
     count ++ 
-    if (count === places.length) {
+    if (count === 10) {
```
EkkiD: What happens the second time you call listResults (after fetching entries 10-19)?

As far as I can tell, placeInfo will still contain the first 10 entries and you're not clearing the innerhtml anymore.

Even if there's no bug here, the fact that this code isn't obvious is a sign that there's room for improvement.

Now that we're clearing the places before sending the place request, would it be possible to just render the places one by one as soon as the responses arrive?

EkkiD (changes requested):

zhoua0805 (approved):

planet/src/main/webapp/places.js
```diff
@@ -241,98 +261,94 @@ function createMarkers(place) {
     });
 }
 
-/** List results of nearby search on page*/
-function listResults() {
+/** List result of searched place on page*/
+function listResults(placeInfo) {
```
EkkiD: nit: this function's name no longer matches its behaviour. It's pluralized, but it's handling a single result.

Maybe `renderResult`?

EkkiD (approved):
LGTM once last comment is addressed

PR #45, opened 2020-07-24
Email verification

nabil-dbz:
- When the user creates an account, a verification email gets sent to their inbox.

- The user gets added to the database, even if the email is not verified, with a property emailVerified that get updated every time they sign in.

-  If the email is not verified, the use can't sign in and their profile page can't be accessed.

### DIFF
--- a/planet/src/main/webapp/js/authentication.js
+++ b/planet/src/main/webapp/js/authentication.js
@@ -44,7 +44,7 @@ document.addEventListener('DOMContentLoaded', function() {
 
 function checkUserSignIn() {
     currentUser = firebase.auth().currentUser;
-    if (currentUser !== null) {
+    if (currentUser !== null && currentUser.emailVerified) {
         document.getElementById('profile-button').innerText = currentUser.displayName;
         document.getElementById('profile-button').style.display = 'block';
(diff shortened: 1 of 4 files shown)


zhoua0805 (approved):
LGTM

planet/src/main/webapp/js/authentication.js
```diff
@@ -86,12 +95,59 @@ async function generateUsername(displayName) {
...
+        addUserToDatabase(user.uid, displayName, phoneNumber, email, 
+            username, user.emailVerified);
+    })
+    .catch(function(error) {
+        const errorCode = error.code;
+        const errorMessage = error.message;
+        if (errorCode === 'auth/missing-android-pkg-name') {
+            alert('Android package name is missing');
+        } else if (errorCode === 'auth/missing-continue-uri') {
+            alert('Continue URL is missing');
```
EkkiD: what is the user supposed to do about these alerts? If there's no user action that can be taken, we can show the user a more friendly generic error and log the real error

nabil-dbz: No! The user can't do anything about them. It's probably better to just log the error on the console

EkkiD (approved):
LGTM

PR #46, opened 2020-07-27
Create load more button

zhoua0805:
- Restructure function renderPlaces() such that it saves all place IDs into a global javascript array
- Create new function makePlaceRequests(), which only makes 10 requests 
- Use global variavle placeCount to keep track of requested places
- Show load more button 5 seconds after makePlaceRequests() is called

**One more issue:**
When 10 requests are made, google maps api returns those 10 place details in random order., so even though we order the placeids by timestamp, the most recently saved place will not be guaranteed render first, but instead anywhere between the first 10 places.

### DIFF
--- a/planet/src/main/webapp/js/placesItinerary.js
+++ b/planet/src/main/webapp/js/placesItinerary.js
@@ -62,10 +62,10 @@ export function renderPlaces() {
     placesRef.orderByValue().once('value', (placesSnapshot) => {
         const placesContainer = document.getElementById('places');
         placesContainer.innerHTML = '';
-        const savedPlaces = placesSnapshot.val();
-        for (const placeId in savedPlaces) {
+        placesSnapshot.forEach(function(childPlace) {
+            const placeId = childPlace.key;
             allPlaces.push(placeId);
-        }
+        });
         makePlaceRequests();
     });
 }
--- a/planet/src/main/webapp/js/placesItinerary.js
+++ b/planet/src/main/webapp/js/placesItinerary.js
@@ -73,10 +73,8 @@ export function renderPlaces() {
 function makePlaceRequests() {
     hideLoadMoreButton();
     let service = new google.maps.places.PlacesService(map);
-    for (let i = placeCount; i < placeCount + 10 ; i++) {
-        if (i >= allPlaces.length) {
-            return;
-        }
+    const maxNumOfRequests = 10;
+    for (let i = placeCount; i < Math.min(placeCount + maxNumOfRequests , allPlaces.length); i++) {
         let placeId = allPlaces[i];
         // Make request with fields
         let placeRequest = {
--- a/planet/src/main/webapp/js/placesItinerary.js
+++ b/planet/src/main/webapp/js/placesItinerary.js
@@ -31,7 +31,7 @@ let toBeProcessedPlaces = []; // Array that contains all the placeIds to be used
 // Declare global variables 
 let autocompleteStart;
 let autocompleteEvent;
-
+ 
 /** Adds autocomplete to input boxes */
 function initAutocomplete() {
     map = new google.maps.Map(document.getElementById('empty-map'));
--- a/planet/src/main/webapp/js/placesItinerary.js
+++ b/planet/src/main/webapp/js/placesItinerary.js
@@ -24,7 +24,9 @@ window.initAutocomplete = initAutocomplete;
 // Declare global variables/constants.
 const database = firebase.database();
 let map;
-let displayedPlaces = {};  // Object that contains all the places information
+let displayedPlaces = {};  // Object that contains all the places information that are rendered
+let allPlaces = []; // Array that contains all the placeIds that the user saved
+let placeCount = 0;
 
 // Declare global variables 
 let autocompleteStart;
@@ -52,23 +54,37 @@ function updateStartingAddress() {
 
 export function renderPlaces() {
     displayedPlaces = {};
+    allPlaces = [];
+    placeCount = 0;
     const userId = firebase.auth().currentUser.uid;
     const placesRef = database.ref('users/' + userId + '/places');
-    placesRef.once('value', (placesSnapshot) => {
+    placesRef.orderByValue().once('value', (placesSnapshot) => {
         const placesContainer = document.getElementById('places');
         placesContainer.innerHTML = '';
         const savedPlaces = placesSnapshot.val();
         for (const placeId in savedPlaces) {
-            // Make request with fields
-            let placeRequest = {
-                placeId: placeId,
-                fields: ['place_id','name','rating','formatted_address',
-                    'opening_hours','photos','url']
-            };
-            let service = new google.maps.places.PlacesService(map);
-            service.getDetails(placeRequest, renderPlaceDetailsCallback); 
+            allPlaces.push(placeId);
         }
+        makePlaceRequests();
     });
+}
+
+function makePlaceRequests() {
+    let service = new google.maps.places.PlacesService(map);
+    for (let i = placeCount; i < placeCount + 10 ; i++) {
+        if (i >= allPlaces.length) {
+            return;
+        }
+        let placeId = allPlaces[i];
+        // Make request with fields
+        let placeRequest = {
+            placeId: placeId,
+            fields: ['place_id','name','rating','formatted_address',
+                'opening_hours','photos','url']
+        };
+        service.getDetails(placeRequest, renderPlaceDetailsCallback);
+    } 
+    setTimeout(renderLoadMoreButton, 5000);
 }
 
 function renderPlaceDetailsCallback(place, status) {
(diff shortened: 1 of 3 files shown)


planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -38,39 +41,75 @@ function initAutocomplete() {
...
+            allPlaces.push(placeId);
         }
+        makePlaceRequests();
     });
 }
 
+function makePlaceRequests() {
+    hideLoadMoreButton();
+    let service = new google.maps.places.PlacesService(map);
+    for (let i = placeCount; i < placeCount + 10 ; i++) {
```
EkkiD: 10 is a magic number here. Prefer adding a CONSTANT above or a local const to indicate what this is.

https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -38,39 +41,75 @@ function initAutocomplete() {
...
+        if (i >= allPlaces.length) {
+            return;
+        }
```
EkkiD: you can probably pull this into the `for` condition `i < min(placeCount + 10, allPlaces.length)`

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -38,39 +41,75 @@ function initAutocomplete() {
...
 function renderPlaceDetailsCallback(place, status) {
     if (status == google.maps.places.PlacesServiceStatus.OK) {
         const placesContainer = document.getElementById('places');
         const placeElement = createPlaceElement(place);
         placesContainer.appendChild(placeElement);
+        placeCount += 1;
     } 
 }
 
+function hideLoadMoreButton() {
```
EkkiD, line 104: +1 for having separate functions for `hide` and `show`

EkkiD (changes requested):

planet/src/main/webapp/js/placesItinerary.js

planet/src/main/webapp/js/placesItinerary.js

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: :D

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -52,33 +55,59 @@ function updateStartingAddress() {
...
+            fields: ['place_id','name','rating','formatted_address',
+                'opening_hours','photos','url']
+        };
+        service.getDetails(placeRequest, renderPlaceDetailsCallback);
+    } 
+    setTimeout(showLoadMoreButton, 5000);
+}
+
 function renderPlaceDetailsCallback(place, status) {
     if (status == google.maps.places.PlacesServiceStatus.OK) {
```
EkkiD, line 92: if one of the requests fails, we'll fetch one of the places twice, right? Is that something we want to worry about?

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -52,33 +55,59 @@ function updateStartingAddress() {
...
+        let placeId = allPlaces[i];
+        // Make request with fields
+        let placeRequest = {
+            placeId: placeId,
+            fields: ['place_id','name','rating','formatted_address',
+                'opening_hours','photos','url']
+        };
+        service.getDetails(placeRequest, renderPlaceDetailsCallback);
+    } 
+    setTimeout(showLoadMoreButton, 5000);
```
EkkiD, line 88: nit: add a comment saying why this is necessary

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -52,33 +55,59 @@ function updateStartingAddress() {
...
         for (const placeId in savedPlaces) {
-            // Make request with fields
-            let placeRequest = {
-                placeId: placeId,
-                fields: ['place_id','name','rating','formatted_address',
-                    'opening_hours','photos','url']
-            };
-            let service = new google.maps.places.PlacesService(map);
-            service.getDetails(placeRequest, renderPlaceDetailsCallback); 
+            allPlaces.push(placeId);
```
EkkiD: Is it possible to just directly assign `allPlaces = savedPlaces` (or copy it with a oneliner, if we need a copy: `allPlaces =  [...savedPlaces]`)

EkkiD (changes requested):

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: `savedPlaces` is a javascript object with placeIds as keys and timestamps as values, however `allPlaces` is an array sorted by timestamp. Therefore i'm not sure if changing allPlaces to objects will work with the existing code.

zhoua0805: Actually, I modified the code a little bit so that the returned data is actually in order (when we call placesSnapshot.val() the order is based on the browser, however if we use a forEach function for the childPlace we can get the correct order)

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: Fixed

planet/src/main/webapp/js/placesItinerary.js
EkkiD: does placesSnapshot.keys() give you the correct ordering?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys

If not, forEach looks good as well.

EkkiD (approved):
LGTM

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: Yes! So the counter will increment by one after each callback. However it seems like the callback functions are very asynchronous and might process some requests earlier than others. 
So i'm not sure how to handle this...

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: Unfortunately no

planet/src/main/webapp/js/placesItinerary.js
EkkiD: Could we keep track of specifically which requests have succeeded and which have not yet succeeded? Then, we could generate new requests based on which have succeeded.

Which data structure do you think we could use for that?

EkkiD (changes requested):

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: Maybe we could have two lists: `allPlaces` and `processedPlaces`. Once a place is successfully requested, move it from allPlaces to processedPlaces?

zhoua0805: Actually we might not need processedPlaces at all, once a place is successfully requested I can remove it from allPlaces

EkkiD: That seems like it would work. Each time you want to request a new batch of places, you can fetch the first 10 from the `unprocessedPlaces` list.

zhoua0805: Fixed :)

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -52,33 +54,65 @@ function updateStartingAddress() {
...
 function renderPlaceDetailsCallback(place, status) {
     if (status == google.maps.places.PlacesServiceStatus.OK) {
         const placesContainer = document.getElementById('places');
         const placeElement = createPlaceElement(place);
         placesContainer.appendChild(placeElement);
+
+        const indexOfProcessedPlace = toBeProcessedPlaces.indexOf(place.place_id);
+        if (indexOfProcessedPlace > -1) {
+            toBeProcessedPlaces.splice(indexOfProcessedPlace, 1);
+        }
```
EkkiD, line 100: nit: using a list makes this code a bit awkward. Consider using a dict instead (in JS that's an object, I think)

EkkiD (changes requested):

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: True. But using an object means we have to keep a counter when looping the first 10 items, which might be slightly awkward too...

EkkiD: I'm not sure I follow why that would be necessary. Using an object for toBeProcessedPlaces would allow you to remove with an easier to understand piece of code:

`delete toBeProcessedPlaces[place.place_id]`

If I'm misunderstanding, feel free to explain here or schedule a quick call for us to discuss

EkkiD (changes requested):

emilyzhaoo (approved):
LGTM

planet/src/main/webapp/js/placesItinerary.js
zhoua0805: haha I was talking about when we need to send another batch of 10 requests, if `toBeProcessedPlaces` is an object, we will need to call `object.keys()` to get an array of placeIds, so that we can loop through the first 10 keys (or a forEach loop with a counter). We'll also need to do the same thing for getting the object size (`object.keys().length`)
The delete function will be easier, but looping the first 10 items will be more complicated

EkkiD: ok it seems like this is a reasonable tradeoff.

EkkiD (approved):
LGTM

PR #48, opened 2020-07-28
Reset Password

nabil-dbz:
- A new dialog gets opened when the user clicks on "Forgot Password?"
- In order to get a reset password email, the user is asked to enter their email address
- If the email gets sent, a success message gets rendered in a green box
- If an error occurs, a message gets rendered in a red box


planet/src/main/webapp/js/authentication.js
```diff
@@ -279,6 +280,52 @@ function signInWithProvider(provider) {
...
+        } else if (errorCode === 'auth/missing-continue-uri') {
+            message.innerText = 'Reset email has not been sent! Please try again!';
+        } else if (errorCode === 'auth/missing-ios-bundle-id') {
+            message.innerText = 'Reset email has not been sent! Please try again!';
+        } else if (errorCode === 'auth/invalid-continue-uri') {
+            message.innerText = 'Reset email has not been sent! Please try again!';
+        } else if (errorCode === 'auth/unauthorized-continue-uri') {
+            message.innerText = 'Reset email has not been sent! Please try again!';
+        } else if (errorCode === 'auth/user-not-found') {
+            message.innerText = 'User not found! Make sure the email is correct!';
```
EkkiD: This is a tricky situation. In theory, we shouldn't show this to the user because it divulges information about our set of users. 

E.g. if I'm malicious and want to know whether or not you have an account on this website, I can now find out.

Normally, we would just tell the user something like:

"If the account was found, an email was sent to user@email.com to reset the password."

Then, we would hide these errors in the server side to prevent leaking any of this info at all. However, I don't think that's possible with firebase, because this info would be present in the response object and the user can inspect it in the chrome dev tools. 

Long story short, I think this is fine given the limitations of the firebase API, but in general we would try to hide this specific error from users.

nabil-dbz: Interesting point! So, if the email has not been sent for any reason, should I just print "If the account was found, an email was sent to user@email.com to reset the password." even though I know it hasn't been sent?

EkkiD: Yes, I think that would be best here.

It doesn't actually add any real security, since the chrome dev tools will show the response, but it's the right philosophy.

EkkiD (approved):
LGTM

planet/src/main/webapp/js/authentication.js
nabil-dbz: Okay I'll change that right away!

emilyzhaoo (approved):
LGTM

PR #49, opened 2020-07-28
Autocomplete search bar

nabil-dbz:
- Move the navigation bar into another file to be loaded when the page gets loaded
- Add a search bar in the navigation bar
- When the user types in the search bar, the list of users whose names start with the input value get rendered (5 users max)
- When the user hovers on a search results element, its background changes

TO DOs:
- Navigate through the list of results using the key arrows (up and down)
- Add a click event handler to the elements


planet/src/main/webapp/navbar.html
```diff
@@ -0,0 +1,22 @@
...
+        <li><a id="profile-button" style="display: none;"></a></li>
+        <li><a id="sign-in-button" style="display: none;" onclick="openModal('sign-in-modal')"><i
+                    class="material-icons right">account_circle</i>Sign In</a></li>
+    </ul>
+    <ul id="search-bar-container" class="right hide-on-med-and-down">
+        <li id="search-bar" class="row">
+            <i id="search-icon" class="col s1 material-icons">search</i>
+            <div class="col s11 input-field">
+                <input id="search-input" type="search" placeholder="Search"
+                    style="height: auto; padding:0; margin: 0;" onkeyup="onKeyUp()"
```
emilyzhaoo: Why is the search bar associated with the up key? Wouldn't we want it to be used with any change? such as  using an event change listener

planet/src/main/webapp/js/search-bar.js
```diff
@@ -0,0 +1,70 @@
...
+
+function hideContainer() {
+    const resultsContainer = document.getElementById('search-results-container');
+    resultsContainer.style.display = 'none';
+    resultsContainer.style.opacity = '0';
+}
+
+function addSearchResultElement(name, username) {
+    const newElement = document.createElement('li');
+
```
emilyzhaoo, line 52: nit: remove extra line

planet/src/main/webapp/js/search-bar.js
```diff
@@ -0,0 +1,70 @@
...
+    nameElement.style.fontSize = 'min(1.7vw, 16px)';
+    nameElement.style.paddingRight = '1%';
+    nameElement.style.margin = '0';
+    
+    const usernameElement = document.createElement('span');
+    usernameElement.innerText = '(' + username + ')';
+    usernameElement.style.fontSize = 'min(1.2vw, 12px)';
+    usernameElement.style.margin = '0';
+    
+
```
emilyzhaoo: nit: remove extra line

emilyzhaoo (changes requested):

planet/src/main/webapp/navbar.html
nabil-dbz: Yeah! What kind of event would you suggest listening to instead of the keyup event?

nabil-dbz: Actually, I just found the oninput event listener that works on the <input> and <select> elements. I will be using it instead on the onkeyup. However, I will still need the onkeyup event to listen to the arrow keys as well as the enter key for the next PR.

emilyzhaoo (approved):
LGTM

planet/src/main/webapp/js/search-bar.js
```diff
@@ -0,0 +1,69 @@
...
+function getSearchResults() {
+    const resultsContainer = document.getElementById('search-results-container');
+    resultsContainer.innerHTML = '';
+    const searchInput = document.getElementById('search-input').value;
+    if (searchInput === '') {
+        return;
+    }
+    database.ref('users').orderByChild('name')  // Order elements by name
+    .startAt(searchInput + 'A') // Start at the users whose names start with searchInput + 'A'
```
EkkiD, line 23: are there any non-alphabetic characters in usernames? If so, you might not catch them here

EkkiD (changes requested):

planet/src/main/webapp/js/search-bar.js
nabil-dbz: True! The only non-alphabetic characters in usernames are the space character! I added more characters in the next PR. Should I also add them here?

planet/src/main/webapp/js/search-bar.js
```diff
@@ -0,0 +1,69 @@
...
+// limitations under the License.
+
+function getSearchResults() {
+    const resultsContainer = document.getElementById('search-results-container');
+    resultsContainer.innerHTML = '';
+    const searchInput = document.getElementById('search-input').value;
+    if (searchInput === '') {
+        return;
+    }
+    database.ref('users').orderByChild('name')  // Order elements by name
```
EkkiD, line 22: do we already have an index on `name`? If not, it might be worth adding it

planet/src/main/webapp/js/search-bar.js
EkkiD: Is space is ordered lexicographically between 'A' and 'z'? I see that it does between those values in the ASCII and Unicode specs, but I don't know where it sorts lexicographically. Have you confirmed this works correctly with spaces?

You can update this logic either in this PR or in the next one to handle other values.

EkkiD (changes requested):

planet/src/main/webapp/js/search-bar.js
nabil-dbz: It tried to but I couldn't since I already have an indexOn rule for the username.

planet/src/main/webapp/js/search-bar.js
nabil-dbz: In the other PR, I have included almost all special characters (from the space character to 'z').

planet/src/main/webapp/js/search-bar.js
EkkiD: (Ignore this if you've already tried it.)

It looks like you can have multiple indexes for a given table
https://firebase.google.com/docs/database/security/indexing-data?hl=en
```
{
  "rules": {
    "dinosaurs": {
      ".indexOn": ["height", "length"]
    }
  }
}
```

planet/src/main/webapp/js/search-bar.js
EkkiD: sounds good.

EkkiD (changes requested):

planet/src/main/webapp/js/search-bar.js
nabil-dbz: Haha it looks like it's working! Done! Thank you

EkkiD (approved):
LGTM

PR #52, opened 2020-07-29
Add UI to allow users to choose  between in order or optimized

zhoua0805:
- Change button to select tag with two options: schedule itinerary in order or schedule optimized itinerary
- Pass boolean `optimized` as a query string

### DIFF
--- a/planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
+++ b/planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
@@ -20,12 +20,8 @@ import com.google.gson.reflect.TypeToken;
 import com.google.planet.data.Event;
 import com.google.planet.data.ItineraryGenerator;
 import com.google.planet.data.ItineraryItem;
-<<<<<<< HEAD
 import com.google.planet.data.ItineraryOrder;
-import com.google.planet.data.Itinerary;
-=======
 import com.google.planet.data.ItineraryException;
->>>>>>> ba2c40a080918f5f34b1f4f49326ce35e9865ae1
 import com.google.planet.data.TimeRange;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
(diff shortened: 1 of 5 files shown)


planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
```diff
@@ -56,7 +58,14 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
             try {
                 ItineraryGenerator itineraryGenerator = new ItineraryGenerator();
-                List<ItineraryItem> itinerary = itineraryGenerator.generateItinerary(events);
+                List<ItineraryItem> itinerary = new ArrayList();
+                if (optimizedString.equals("true")){
+                    itinerary = itineraryGenerator.generateItinerary(events, ItineraryOrder.OPTIMIZED);
+                } else if (optimizedString.equals("false")){
+                    itinerary = itineraryGenerator.generateItinerary(events, ItineraryOrder.UNOPTIMIZED);
+                } else {
+                    throw new ItineraryException("Optimized can only be true or false");
```
EkkiD, line 67: if it's empty string, would we want to default to something?

EkkiD (changes requested):

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
zhoua0805: Technically it would never be anything other than true/false unless the user changes something in the dev tools console. And if that's the case I think it would be okay to skip the default

planet/src/main/webapp/itinerary.html
```diff
@@ -92,7 +92,11 @@ <h4> Step 2: Create events </h4>
                     <button class="btn waves-effect waves-light" onclick="saveEvents();"><i class="material-icons">cloud_done</i></button>
                     <button class="btn waves-effect waves-light" onclick="closeSaveEventsForm();"><i class="material-icons">highlight_off</i></button></br>
                 </div>
-                <button class="btn waves-effect waves-light" onclick="generateItinerary()">generage itnerary</button>
+                <select id="itinerary-selection" class="browser-default btn waves-effect waves-light" onchange="handleItinerarySelectionChange();">
+                    <option id="default-itinerary-selection" value="" disabled selected>generage itnerary</option>
+                    <option value="true">generate optimized itinerary</option>
+                    <option value="false">generate itinerary in order</option>
+                </select>
```
emilyzhaoo: nit: extra line

emilyzhaoo (approved):
LGTM

EkkiD (approved):
LGTM

PR #54, opened 2020-07-30
Test generating itinerary in order and generating optimized itinerary

zhoua0805:
- Add two test cases to test for generating itinerary in order and generating optimized itinerary
- Use mockito to mock the function getDirectionRoute which uses the directions api
- TODO: test more edge cases such as no events, too many events, etc.

### DIFF
--- a/planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
+++ b/planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
@@ -29,7 +29,7 @@ import com.google.maps.model.DirectionsRoute;
 import com.google.maps.model.DirectionsLeg;
 import com.google.maps.model.Duration;
 
-public final class ItineraryGenerator {
+public class ItineraryGenerator {
     public List<ItineraryItem> generateItinerary(List<Event> events) throws ItineraryException{
 
         // TODO: This needs to be removed in the next PR and changed to a function parameter
@@ -119,7 +119,7 @@ public final class ItineraryGenerator {
     // DirectionRoute also has int[] waypointOrder, which will be empty if optimized is set to false,
     // and will contain the new order of waypoints if optimized is set to true
     // For the MVP, real time traffic is NOT used.
-    private DirectionsRoute getDirectionsRoute(List<Event> events, boolean optimized) throws ItineraryException{
+    public DirectionsRoute getDirectionsRoute(List<Event> events, boolean optimized) throws ItineraryException{
         String GoogleApiKey = "AIzaSyDK36gDoYgOj4AlbCqh1IuaUuTlcpKF0ns";
         String origin = events.get(0).getAddress();
         String destination = origin; // The ending location should be assumed as the starting location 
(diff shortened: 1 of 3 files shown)


planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
```diff
@@ -84,7 +84,7 @@
     // Each DirectionsLeg contains starting address, ending address, and duration
     // DirectionRoute also has int[] waypointOrder, which will contain the order of the waypoints
     // For the MVP, real time traffic is NOT used.
-    private DirectionsRoute getDirectionsRoute(List<Event> events, boolean optimized) throws ItineraryException{
+    public DirectionsRoute getDirectionsRoute(List<Event> events, boolean optimized) throws ItineraryException{
```
azhang72: Instead of making a method public just so that its own test can mock it, we can have a helper class as a thin wrapper around Maps API. We pass the helper class into ItineraryGenerator in its constructor and mock the helper in tests.
The pattern is called [_dependency injection_](https://en.wikipedia.org/wiki/Dependency_injection) in case you want to read more about it.
I recommend you refactoring the code if you have time in your last week.

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
```diff
@@ -20,8 +20,8 @@
 import com.google.planet.data.Event;
 import com.google.planet.data.ItineraryGenerator;
 import com.google.planet.data.ItineraryItem;
-import com.google.planet.data.ItineraryOrder;
 import com.google.planet.data.ItineraryException;
+import com.google.planet.data.ItineraryOrder;
```
azhang72, line 24: Do we need this?

planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
zhoua0805: Ah okay, thank you!

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
zhoua0805: ItineraryOrder has OPTIMIZED and UNOPTIMIZED which are boolean values to pass in the itinerary generator function. I used them instead of primitives (true/false) hoping to make the code more readable

azhang72: Why is the import line the only change in this file? Did you start using the enums in a previous change and it wouldn't build?

planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
azhang72: Also consider replace the boolean |optimized| with your enum in this file

planet/src/main/java/com/google/planet/servlets/ItineraryServlet.java
zhoua0805: Oh it was there before too hahah. I accidentally reordered it when resolving merge conflicts

planet/src/main/java/com/google/planet/data/ItineraryGenerator.java
zhoua0805: Fixed!

azhang72 (approved):

emilyzhaoo (approved):
LGTM

PR #56, opened 2020-07-31
Display Saved Icons on Refresh

emilyzhaoo:
Display previously saved places as "Saved" after refresh of page.

Create materialize Toast message when user attempts to click save icon when not signed in. Toast message tells user to sign in to start saving.Save icons appear as saved after refresh of page if the place is saved inside the database.


azhang72 (approved):

planet/src/main/webapp/places.js
```diff
@@ -353,10 +353,21 @@ function renderResult(placeInfo) {
...
 
+/** Create and add save icon */
 function createIcon(placeID) {
-    // Create and add save icon 
     let icon = document.createElement('i');
-    icon.innerHTML = 'favorite_border';
+    let userID = currentUser.uid; 
+    
+    database.ref('users/' + userID +'/places').child(placeID).once('value').then(function(snapshot) {
+        // Check if place is already saved in database
```
EkkiD: This comment is just repeating the code. It doesn't help the user understand.

Instead, we can explain _why_ the fact that it's in the database affects the styling.

E.g.
// Places are in the database if and only if they've been favourited by the user.

EkkiD (changes requested):

EkkiD (approved):
LGTM

PR #57, opened 2020-08-04
Add date to the list of events

nabil-dbz:
**NOTE: @zhoua0805 has to review this PR before it gets merged!**

- When the user saves the list of events, they have to add the date
- The date has to be valid
- The function saveEvents() adds the user to the list of visitors in the places data tree
- When the event gets deleted, the user is deleted from the place's list of visitors of that day only
- The places data tree follows this structure:
Screenshot 2020-08-04 at 10 39 20 AM

**TODO:**
- The time this PR is saving is not the time when the user is visiting the place since the itinerary is not yet generated

### DIFF
--- a/planet/src/main/webapp/js/itinerary.js
+++ b/planet/src/main/webapp/js/itinerary.js
@@ -192,13 +192,15 @@ function renderEvents(listName) {
         const eventsContainer = document.getElementById('events');
         eventsContainer.innerHTML = '';
         eventListSnapshot.forEach(function(childEvent) {
-            let eventObject = childEvent.val();
-            let eventElement = createEventElement (listName,
-                                                childEvent.key,
-                                                eventObject.name, 
-                                                eventObject.address, 
-                                                eventObject.duration);
-            eventsContainer.appendChild(eventElement);
+            if (childEvent.key !== 'date') {
+                let eventObject = childEvent.val();
+                let eventElement = createEventElement (listName,
+                                                    childEvent.key,
+                                                    eventObject.name, 
+                                                    eventObject.address, 
+                                                    eventObject.duration);
+                eventsContainer.appendChild(eventElement);
+            }
         });
     });
 }
@@ -228,27 +230,30 @@ async function deleteEvent(listName, ref) {
     const userId = firebase.auth().currentUser.uid;
     const eventListRef = database.ref('events/' + userId + '/' + listName);
     const toBeDeletedEventRef = eventListRef.child(ref);
-    const toBeDeletedEventSnap = await toBeDeletedEventRef.once('value');
-    const eventDate = toBeDeletedEventSnap.val()['date'];
+    let eventDate = '08/07/2020'; // Default value
     toBeDeletedEventRef.remove();
 
     // Fix order after deleting the event
     eventListRef.orderByChild('order').once('value', (eventListSnapshot) => {
         let count = 1;
         eventListSnapshot.forEach(function(childEvent) {
-            let event = childEvent.val();
-            let eventKey = childEvent.key;
-            if (event.order !== count) {
-                eventListRef.child(eventKey).update({
-                    order: count
-                });
+            if (childEvent.key !== 'date'){
+                let event = childEvent.val();
+                let eventKey = childEvent.key;
+                if (event.order !== count) {
+                    eventListRef.child(eventKey).update({
+                        order: count
+                    });
+                }
+                count += 1;
+            } else {
+                eventDate = childEvent.val();
             }
-            count += 1;
         });
     });
 
     if (listName !== 'currentList') {
-        removeVisitor(userId, ref, eventDate);
+        deleteVisitor(userId, ref, eventDate);
     }
 
     // Render places again since the icons might need to change
@@ -271,7 +276,10 @@ async function saveEvents() {
         return;
     }
     newListRef.set(currentListSnapshot.val());
-    addUserToAllPlaces(userId, newListRef, listDate);
+    newListRef.update({
+        date: listDate
+    });
+    addUserToAllPlaces(userId, currentListSnapshot, listDate);
     // Clear current list
     currentListRef.remove();
     // Update the select tag options and close save events form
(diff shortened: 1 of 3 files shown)


emilyzhaoo (approved):
lgtm

planet/src/main/webapp/js/itinerary.js
```diff
@@ -224,44 +226,50 @@ function createEventElement(listName, ref, name, address, duration) {
...
-    // Check if this is an event created from a place, and if so, 
-    // remove the user id from the visitors list 
-    const placeRef = database.ref('places/' + ref);
-    placeRef.once('value', (placeSnapshot) => {
-        if (placeSnapshot.val()) {
-            placeRef.child('visitors').child(userId).remove();
-        }
-    });
+    // We check that the list is saved by the user
+    // If the user saved the list than the listDate is not the default value
```
zhoua0805: nit: typo than -> then

planet/src/main/webapp/js/itinerary.js
```diff
@@ -281,6 +293,40 @@ async function saveEvents() {
     renderPlaceButtons();
 }
 
+function isDateValid(date) {
+    return Date.parse(date) > Date.now();
+}
+
+async function addUserToAllPlaces(userId, eventsSnapshot, date) {
+    // TODO: Consider the real time range of the events
+    let time = 600;
```
zhoua0805, line 304: What does the variable time represent here? Is it the start time of each event? Is this something we want to show on the profile page? If so, would people know the exact time you're going to an event?

planet/src/main/webapp/js/itinerary.js
```diff
@@ -224,44 +226,50 @@ function createEventElement(listName, ref, name, address, duration) {
...
-    // remove the user id from the visitors list 
-    const placeRef = database.ref('places/' + ref);
-    placeRef.once('value', (placeSnapshot) => {
-        if (placeSnapshot.val()) {
-            placeRef.child('visitors').child(userId).remove();
-        }
-    });
+    // We check that the list is saved by the user
+    // If the user saved the list than the listDate is not the default value
+    if (listDate !== '08-07-2020') {
```
azhang72: Save '08-07-2020' in a constant.

azhang72 (changes requested):
What is the significance of this date you are saving if it's not the same as the visit date in itinerary?

planet/src/main/webapp/js/itinerary.js
nabil-dbz: Actually! No, we are not going to show this info on the profile page but it would be interesting to filter the list of visitors by the time of the visit which is something I doubt we will be able to implement this week but we'll filter by day only.

nabil-dbz: Check the PR's description: the time we are storing the database is not really the time of the visit because the itinerary is yet to be generated.

planet/src/main/webapp/js/itinerary.js
nabil-dbz: Yeah I though the comment would do, but I'll save it in a constant to make sure is more understandable.

planet/src/main/webapp/js/itinerary.js
nabil-dbz: Done!

planet/src/main/webapp/js/itinerary.js
zhoua0805: Ah okay!

zhoua0805 (approved):
LGTM

planet/src/main/webapp/js/itinerary.js
azhang72: Please also address my other comment on the significance of this date.

nabil-dbz: Oh! Sorry didn't see that! 
Actually, supposedly 08-07-2020 is the date we release the application (the presentation day), so we are sure that all the saved lists will have a date different than this because they have to be saved for after this date to be valid.

azhang72: What is the meaning of the |listDate| if it's not the one when the users visit a certain place?

nabil-dbz: It is the date when the users visit a certain place, but it's possible for us not to find this information in the database (for the currentList), that's why I am giving it a default value.

azhang72: Add the TODO in your PR message to the code.

planet/src/main/webapp/js/placesItinerary.js
```diff
@@ -254,16 +254,6 @@ async function submitPlace(placeId) {
         order: order,
     });
```
zhoua0805, line 256: We will need to call addVisitor here if the user is making edits on a saved list (i.e not currentList)

zhoua0805 (changes requested):

planet/src/main/webapp/js/itinerary.js
nabil-dbz: It's there ... Line 302 of the itinerary.js file

planet/src/main/webapp/js/itinerary.js
```diff
@@ -224,44 +226,51 @@ function createEventElement(listName, ref, name, address, duration) {
...
-    // remove the user id from the visitors list 
-    const placeRef = database.ref('places/' + ref);
-    placeRef.once('value', (placeSnapshot) => {
-        if (placeSnapshot.val()) {
-            placeRef.child('visitors').child(userId).remove();
-        }
-    });
+    // We check that the list is saved by the user
+    // If the user saved the list then the listDate is not the default value
+    if (listDate !== DEFAULT_DATE) {
```
zhoua0805, line 258: The alternative would be directly checking if listName === 'currentList'

azhang72 (approved):

planet/src/main/webapp/js/itinerary.js
nabil-dbz: I went for idea that at first but then I wasn't sure if you handled the case where the user names the list they want to save "currentList"

zhoua0805: Good question. I didn't handle that. Maybe we can add an error checking preventing users to name their list as "currentList".

nabil-dbz: Yeah that would be a good idea!

planet/src/main/webapp/js/placesItinerary.js
nabil-dbz: Done!

zhoua0805 (changes requested):
Don't forget to skip the date in generateItinerary() function

zhoua0805 (approved):
LGTM

PR #58, opened 2020-08-04
Navigate the search results list and access the profile

nabil-dbz:
- The user is able to navigate through the list of elements using the mouse or the keyboard
- The element highlighted, background lightgray, is the one to be access when the enter key is pressed
- Add the user's profile picture in the element
- The user can use the key arrows to go through the list of elements
- The maximum number of elements in the result list is 5 (can be changed)


planet/src/main/webapp/js/search-bar.js
```diff
@@ -12,29 +12,76 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+const UP_ARROW = 38;
+const DOWN_ARROW = 40;
+const ENTER_KEY = 13;
```
azhang72: Where are these key codes defined?

nabil-dbz: In the search-bar script as global constants. Should I put them inside the function or in a separate file?

azhang72: I assume you got these key codes from some standard and didn't come up with the values yourself? Where is that standard and can you directly reference the key values from there? That way, your keyboard navigation won't break if someday the key values change.

nabil-dbz: I actually got the keycodes form this page: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode
I just realized that it's inadvisable so I'll go for another simpler way: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

nabil-dbz: Done!

EkkiD (approved):
LGTM

azhang72 (approved):

emilyzhaoo (approved):
LGTM

PR #61, opened 2020-08-05
Set up mail api

zhoua0805:
**NOTE**: Mail API only works in Java 8 run time

- Send a mail with attachment to a fixed email address

TODO:
 - the pdf attachment was not able be downloaded, need to find a fix
 - send to user's email address
 - send the actual itinerary
   - send from html or call itinerary generator again?
 - hide frontend button when itinerary are not generated yet


planet/src/main/java/com/google/planet/servlets/SendEmailServlet.java
```diff
@@ -0,0 +1,80 @@
...
+                "<h3>You can see a copy of your itinerary in the attachment below.</h3></div>";  
+            htmlPart.setContent(htmlBody, "text/html");
+            multipartContent.addBodyPart(htmlPart);
+
+            // Add a pdf attachment of the itinerary
+            // TODO: change input to actual itinerary
+            MimeBodyPart attachment = new MimeBodyPart();
+            String itinerary = "your itinerary";
+            InputStream attachmentDataStream = new ByteArrayInputStream(itinerary.getBytes("UTF-16"));
+            attachment.setFileName("itinerary.pdf");
```
azhang72, line 66: nit: might want to append a time stamp to the filename in case a user generates more than 1 itinerary.

azhang72 (approved):
LGTM with a nit. I'm curious about the unable to download PDF error if you'd like to discuss it further.

planet/src/main/java/com/google/planet/servlets/SendEmailServlet.java
```diff
@@ -0,0 +1,80 @@
...
+            String htmlBody = "<div>Hello! <h3> Thank you for using Planet.</h3> " +
+                "<h3>You can see a copy of your itinerary in the attachment below.</h3></div>";  
+            htmlPart.setContent(htmlBody, "text/html");
+            multipartContent.addBodyPart(htmlPart);
+
+            // Add a pdf attachment of the itinerary
+            // TODO: change input to actual itinerary
+            MimeBodyPart attachment = new MimeBodyPart();
+            String itinerary = "your itinerary";
+            InputStream attachmentDataStream = new ByteArrayInputStream(itinerary.getBytes("UTF-16"));
```
EkkiD, line 65: looks like you're taking a string and putting it into the stream. That will give us a text file.

I don't know whether a raw text file is a valid PDF, but I would consider setting the content type to `text/plain` and seeing what happens.

EkkiD: Here's an example of an actual minimal PDF file if you were to hand code it: https://brendanzagaeski.appspot.com/0004.html

Generally, I would recommend constructing either a text file and sending that, or constructing an html file and using an html to pdf converter.

EkkiD (changes requested):

nabil-dbz (approved):

planet/src/main/java/com/google/planet/servlets/SendEmailServlet.java
zhoua0805: txt file works! I think I'll go for this solution since it sounds pretty straight forward

planet/src/main/java/com/google/planet/servlets/SendEmailServlet.java
zhoua0805: Okay!

EkkiD (approved):
LGTM

PR #62, opened 2020-08-05
Implement Blobstore Api

nabil-dbz:
This pull request implements the Blobstore Api:
- When the user opens the modal to update the profile picture, the UploadUrlServlet gets called to create an upload url
- When the user chooses the picture and confirms, the UploadFileServlet gets fetched to generate a blobKey
- The UploadFileServlet redirects to the UserServlet by putting the blobKey as a query string
- The UserServlet forwards the blobKey to the profile script which fetchs the ServeFileServlet to get the image and store it in the database

### DIFF
--- a/planet/src/main/java/com/google/planet/servlets/ServeFileServlet.java
+++ b/planet/src/main/java/com/google/planet/servlets/ServeFileServlet.java
@@ -1,0 +1,37 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.planet.servlets;
+
+import java.io.IOException;
+
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import com.google.appengine.api.blobstore.BlobKey;
+import com.google.appengine.api.blobstore.BlobstoreService;
+import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+
+@WebServlet("/serve")
+public class ServeFileServlet extends HttpServlet {
+    private BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+    
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        BlobKey blobKey = new BlobKey(request.getParameter("blob-key"));
+        blobstoreService.serve(blobKey, response);
+    }
+}
--- a/planet/src/main/java/com/google/planet/servlets/ServeFileServlet.java
+++ b/planet/src/main/java/com/google/planet/servlets/ServeFileServlet.java
@@ -25,7 +25,7 @@ import com.google.appengine.api.blobstore.BlobKey;
 import com.google.appengine.api.blobstore.BlobstoreService;
 import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
 
-@WebServlet("/serve")
+@WebServlet("/serve-file")
 public class ServeFileServlet extends HttpServlet {
     private BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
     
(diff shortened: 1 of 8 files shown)


planet/src/main/java/com/google/planet/servlets/ServeFileServlet.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import com.google.appengine.api.blobstore.BlobKey;
+import com.google.appengine.api.blobstore.BlobstoreService;
+import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+
+@WebServlet("/serve")
```
EkkiD: this name is at the top level for all of planet. In that context, "serve" isn't very descriptive. Maybe 'serveblobstore' or 'getimage' or 'serveimage', or something like that

EkkiD: or "serve-file" to match your "upload-file" path below

planet/src/main/java/com/google/planet/servlets/UploadUrlServlet.java
```diff
@@ -0,0 +1,41 @@
...
+@WebServlet("blobstore-upload-url")
+public class UploadUrlServlet extends HttpServlet {
+
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        String username = request.getParameter("username");
+        BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+        String uploadUrl = blobstoreService.createUploadUrl("/upload-file?username=" + username);
```
EkkiD, line 37: nit: add a comment that the passed parameter will be called by blobstore after the upload is complete.

planet/src/main/java/com/google/planet/servlets/UploadFileServlet.java
```diff
@@ -0,0 +1,48 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import com.google.appengine.api.blobstore.BlobKey;
+import com.google.appengine.api.blobstore.BlobstoreService;
+import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+
+@WebServlet("/upload-file")
+public class UploadFileServlet extends HttpServlet {
```
EkkiD, line 32: nit: add a comment explaining how this is called (if I understand correctly, it's called by Blobstore after the user completes the upload?)

planet/src/main/webapp/js/profile.js
```diff
@@ -100,30 +101,35 @@ function switchSection(linkId, sectionId) {
...
-    
-        if (user['emailVerified']) {
+        // If the blobkey exists, get the image url and store in the database
+        if (blobKey !== 'null') {
+            let picture = await fetch('/serve?blob-key=' + blobKey);
+            database.ref('/users/' + userId).update({
+                profilePic: picture.url
+            })
+            .then(() => {
+                window.location.href = '/user/' + username;
```
EkkiD: Why do we need to redirect after this DB write?

EkkiD (changes requested):

planet/src/main/webapp/js/profile.js
nabil-dbz: Actually, it's just to hide the blobkey from the user. I don't know if it's important but if we don't redirect the user can see the blobkey from the profile's URL. Do you think I should remove it?

planet/src/main/java/com/google/planet/servlets/UploadFileServlet.java
nabil-dbz: Done!

planet/src/main/java/com/google/planet/servlets/UploadUrlServlet.java
nabil-dbz: Done!

planet/src/main/java/com/google/planet/servlets/ServeFileServlet.java
nabil-dbz: Done!

zhoua0805 (approved):
LGTM

planet/src/main/webapp/js/profile.js
EkkiD: What does the blob key give the user access to? Does it let them just download their own profile pic? If so, it should be fine to leave it.

Another consideration would be whether or not this is noticeable for the user. Does this line cause the whole page to reload or does the browser just change the URL bar but not reload anything?

nabil-dbz: It actually reloads the page that's why I didn't like this solution much. As for the info in the blobkey, it's just the blobkey of the user's profile picture. I will remove it then and keep the URL with the blobkey!

EkkiD (approved):
LGTM

PR #66, opened 2020-08-06
Render the list of visitors

nabil-dbz:
- The user is able to see the date of the saved list under its name on the profile page
- When the user clicks on the people's icon, the list of the user visiting the same place at the same day gets displayed
- The user can change the date to see the list of visitors of the same place in other days
- The list item in the list of visitors contains the name, username, and profile picture of the user
- If no one is visiting the place, a message gets displayed to inform the user
visitors

### DIFF
--- a/planet/src/main/webapp/js/profile-events-renderer.js
+++ b/planet/src/main/webapp/js/profile-events-renderer.js
@@ -133,8 +133,13 @@ function createEvent(name, address, duration, eventId) {
     usersIcon.style.right = '1%';
     usersIcon.innerText = 'people';
     usersIcon.title = 'People who visited the place';
-    usersIcon.addEventListener('click', function() { 
-        renderVisitorsList(event.target.id);
+    usersIcon.addEventListener('click', function() {
+        const modalElement = document.getElementById('list-visitors');
+        const dateContainer = document.getElementById('date-container');
+        dateContainer.innerHTML = '';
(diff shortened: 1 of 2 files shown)


planet/src/main/webapp/js/profile-events-renderer.js
```diff
@@ -68,13 +69,18 @@ function getListFromSnapshot(index, snapshot) {
 function getListFromObject(object) {
     let list = [];
     for (const property in object) {
-        object[property].id = property;
-        list.push(object[property]);
+        // Within the list of events, we have the date property
+        if (property !== 'date') {
+            object[property].id = property;
```
azhang72: What is this line doing?

nabil-dbz: It's adding the id of the event to the object (event) which can then be used to reference the event ... The event as it is in the database doesn't have a property id so I needed to save it!

zhoua0805 (approved):
LGTM

planet/src/main/webapp/js/profile-events-renderer.js
azhang72: What type is |object| supposed to be here?

nabil-dbz: |object| stands for the list of events snapshot so it's an object. I have just changed the names of variables and functions to make more readable!

azhang72 (approved):

